PR #67, opened 2019-12-18
Fix docker-compose port specification on modern versions of docker.

michaelfavia:
Thanks for the excellent work on this project. This was required to get the containers to launch correctly on my version of docker for mac downloaded today at docker hub. I haven't tested this across other platforms but without it it was complaining about already used ports. 

For example: 
ERROR: for starter_dev_1  Cannot start service dev: driver failed programming external connectivity on endpoint starter_dev_1 (86c8344f8e3db2753ae51c99d7cc9a7b1da891d85bb896d54819b0669820629b): Bind for 0.0.0.0:9757 failed: port is already allocated

ERROR: for dev  Cannot start service dev: driver failed programming external connectivity on endpoint starter_dev_1 (86c8344f8e3db2753ae51c99d7cc9a7b1da891d85bb896d54819b0669820629b): Bind for 0.0.0.0:9757 failed: port is already allocated

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -99,11 +99,11 @@ services:
       - default
     ports:
       # The main server
-      - "5678:5678"
+      - "5678"
       # Node `--inspect=...` port numbers for debugging
-      - "9678:9678"
-      - "9876:9876"
-      - "9757:9757"
+      - "9678"
+      - "9876"
+      - "9757"
     depends_on:
       - db
     # Overrides default command so things don't shut down after the process ends.


charlex:
I ran into this issue too, but it was pretty sporadic, and fairly uncommon.

If you simply quit and reopen docker, does it work without this change?

benjie:
Does this definitely fix an issue? Could you point me to the relevant documentation where it indicates this is desired?

keithlayne:
I don't think this is actually relevant. I too have seen this error, and it seems to be from docker getting hung.  Restarting the docker service has always solved this for me.  

See https://docs.docker.com/compose/compose-file/#ports - this change would cause the container ports to be exposed on ephemeral, non-stable ports, which kind of defeats the intent of exposing well-known ports on the host.

keithlayne:
To be clear, I think this probably only affects Mac and Windows, as they run docker in a VM, which is apparently not perfect still.

benjie:
There'd be little point exposing them on ephemeral ports; you'd be better off just not exposing them at all. I'm going to close this, but if anyone has ideas on how to fix this in Docker whilst still enabling the debugging flow I'm all ears!

PR #77, opened 2020-01-01
Update production.Dockerfile

makoncline:
Add /worker/templates/ to copy commands.


benjie:
I expect tests are failing because prettier is failing on master. I want to make it autofix from a GitHub action but haven’t had time to set it up yet.

benjie:
Thanks!

PR #79, opened 2020-01-02
postgraphile 4.6.0-alpha.0

singingwolfboy:


benjie:
Thanks; there's also the https://github.com/graphile/starter/tree/postgraphile-46 branch which does it in a slightly different way (avoiding the need for `!`). I don't plan to merge 4.6 at this time, so I'm going to close this to keep things tidy.

PR #80, opened 2020-01-04
Add Variants section & CRA edition

### LINKED ISSUE #66: CRA for apps without SSR
Having ```@app/client-cra``` for a Create-React-App based version of the ```@app/client``` might be a better option for apps that sit behind the login wall and don't need any SSR.

alexk111:
Closes #66


benjie:
Not sure what happened there.

PR #86, opened 2020-01-04
Faster second Docker build

benjie:
Previously it was still running the `yarn` command again, by moving `ARG` after this we can reuse that build step.


benjie:
Despite this, building in GitHub seems to be very slow for some reason. Perhaps writing the Docker images themselves is slow...? Might make sense to optimise all the `ENV` lines to be on one line, as GitHub actions seems to take 4-5 seconds on each on (?!!)

PR #89, opened 2020-01-10
Fix user forgot password task

benjie:
Fixes #88 

Thanks to @palamethos for raising this issue.

### DIFF
--- a/@app/worker/src/tasks/user__forgot_password.ts
+++ b/@app/worker/src/tasks/user__forgot_password.ts
@@ -3,7 +3,7 @@ import { SendEmailPayload } from "./send_email";
 
 interface UserForgotPasswordPayload {
   /**
-   * user_email id
+   * user id
    */
   id: number;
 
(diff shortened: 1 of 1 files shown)


benjie:
I've evaluated this issue for potential security issues, but I cannot determine any exploitable security issues here - it's just a bug.

Reasoning: this task incorrectly interprets a user ID to be a user_email ID. It then uses the owner-level database client to get the user_email which has this ID (the id of the initiating user), and then from that, determins the user that owns this user_email. The only feature of the user retrieved that is used is the user id, which is not privileged information. The initiating user is then emailed a URL that contains a token belonging to the initiating user and the ID of the user that owns the email address with the same id as their user ID. The total information disclosure here is that there does or does not exist an email address that has the id the same as the current user's ID. This information is not actionable, and really is pretty boring because in the vast majority of cases an email with that id will exist since every user has emails so the max email id will be greater than the max user id.

PR #90, opened 2020-01-10
Fewer steps for faster CI?

benjie:
Maybe this is faster on CI? I noticed even the `ENV` commands were taking a few seconds each there.


benjie:
Meh

Issue #16, opened 2019-10-08
Password strength indicator required


benjie:

MobliMic:
I have started doing some work on this issue. Is this the sort of thing you are looking for?

![image](https://user-images.githubusercontent.com/4713282/71271091-0ca6d180-234b-11ea-91d2-5e315fd237dd.png)

benjie:
Nice; yeah this is the kind of thing I had in mind! Could you also rename "password" to "passphrase" while you're at it? I should try and ban the word `password` from my vocabulary.

MobliMic:
Opened PR #91

benjie:
Implemented in #91

PR #91, opened 2020-01-10
#16 - Adding password strength indicator

MobliMic:
Adding a password strength indicator to the register page and copied over to the security page too


MobliMic:
Forgot to add the passphrase change

MobliMic:
@benjie Tried rebasing the branch but apparently there is still a conflict

PR #92, opened 2020-01-17
Fixes error message not aligning with rest of form

JoeSchr:
Hi, just was taking a look into converting the react register form to vuetify and this misaligned error bugged me.

Before:
image

After:
image


benjie:
Hah; great spot! I’ll see if we can replace it with whitespace of some kind 👍

PR #94, opened 2020-01-18
Next.js 9.2

singingwolfboy:
https://nextjs.org/blog/next-9-2

I wanted to take advantage of the new built-in CSS import support, but it seems we can't. Antd uses LESS for customizing the theme, and in order to make Next.js work with LESS, we need the `@zeit/next-less` package, which needs the `@zeit/next-css` package, which disables the new built-in CSS import support (for backwards compatibility).

We still get improved code-splitting, though!


benjie:
That’s a shame, I’m a fan of CSS modules.

benjie:
It looks like this gas caused some duplicates in yarn.lock; please run `npx yarn-deduplicate && yarn`

singingwolfboy:
Done!

As for CSS modules, we can use them if we drop LESS support in Next.js. But that means we would have to precompile the Antd theme somehow before calling Next, which feels messy and complicated.

PR #100, opened 2020-01-22
Change word "login" to "Sign in"

connorholyday:
The top-right of the page says "Sign in" instead of the word "login" as suggested in the copy.

image


benjie:
Let me just do a quick `prettier` impersonation...

benjie:
Alas, my impersonation was insufficient.

connorholyday:
It's the thought that counts 😉

benjie:
Turns out I didn't trim the trailing space.

PR #101, opened 2020-01-23
Add avatar photo upload component

makoncline:
User can upload avatar image
File is checked for type and size
File is uploaded to AWS S3 bucket using pre signed url
File is deleted with pre signed url
NEXT API route used for pre signing urls
Bucket has public read access for viewing
 
not many changes needed to allow upload of multiple user photos. 
(don't limit the file list to a single value and keep displaying upload button)

There are MANY things that can be improved, but it's what i've got.
image
image
image


benjie:
Thanks for this! It's going to take me a while to getting around to improving it/merging it, but in the mean time it acts as a great demonstration of how to accomplish this to other users :+1: 

You need to add `@app/graphql` as a dependency to `@app/components/package.json`; I believe this is why CI is failing.

benjie:
```
$ prettier --ignore-path .eslintignore --check '**/*.{js,jsx,ts,tsx,graphql,md}'
Checking formatting...
README.md
Code style issues found in the above file(s). Forgot to run Prettier?
```

^ To solve this, `yarn lint:fix` (yes, we even format markdown! :grin:)

@app/client/src/pages/api/s3.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const putParams = {
+      ...params,
+      Expires: 5 * 60,
+    };
+
+    client.getSignedUrl("putObject", putParams, (err, url) => {
+      if (err) {
+        res.json({ success: false, err });
+      } else {
+        res.json({ success: true, url });
```
singingwolfboy: Rather than using a `success` key in the JSON payload, it's better to use the HTTP status code to communicate an error. In this case, if the error is coming from S3, that indicates the status "502 Bad Gateway" because our server is acting as a gateway (or proxy) to S3.

```js
if (err) {
  res.status(504).json({ err });
} else {
  res.json({ url });
}
```

@app/components/src/AvatarUpload.tsx
```diff
@@ -0,0 +1,208 @@
...
+import { UploadChangeParam } from "antd/lib/upload";
+import { UploadFile, RcCustomRequestOptions } from "antd/lib/upload/interface";
+import axios from "axios";
+import {
+  useChangeAvatarMutation,
+  ProfileSettingsForm_UserFragment,
+} from "@app/graphql";
+import { ApolloError } from "apollo-client";
+
+export function slugify(string: string) {
```
singingwolfboy: Instead of defining your own function to make a slug, it's probably better to use an existing package. I just searched through `yarn.lock`, and it looks like we already depends on a library called [`unique-slug`](https://github.com/npm/unique-slug), which would probably suit your needs just fine.

@app/components/src/AvatarUpload.tsx
```diff
@@ -0,0 +1,208 @@
...
+  };
+
+  const onRemove = async () => {
+    if (await deleteUserAvatarFromBucket()) {
+      setFileList(null);
+    }
+  };
+
+  const uploadButton = (
+    <div>
```
singingwolfboy: For accessibility purposes, this should use a `<button>` element. It should probably also have `aria-label` set to something like "Upload avatar". You can use the [Button component in antd](https://ant.design/components/button/).

@app/client/src/next.config.js
```diff
@@ -31,6 +38,12 @@ if (!ROOT_URL) {
       withCss,
       withLess
     )({
+      serverRuntimeConfig: {
+        BUCKET: BUCKET,
+        AWSACCESSKEYID: AWSACCESSKEYID,
+        AWSSECRETKEY: AWSSECRETKEY,
```
singingwolfboy: This is extremely insecure: you are passing the AWS secret key to the browser! That means any user of your application can find your secret key, and make AWS API requests on your behalf! 😱 

A better, more secure way of doing this is to generate a signed upload URL on the server, and only pass that signed URL to the client. That way, the client has enough information to do what it needs to do (upload files to S3), but nothing more than that. Fortunately, there is a library called [`react-s3-uploader`](https://github.com/odysseyscience/react-s3-uploader) that is designed to work in exactly this way. I highly suggest using this library.

@app/components/src/AvatarUpload.tsx
```diff
@@ -0,0 +1,208 @@
...
+  //   setFileList([...fileList]);
+  // };
+
+  const beforeUpload = (file: any) => {
+    const fileName = file.name.split(".")[0];
+    const fileType = file.name.split(".")[1];
+    file.uid = getUid(fileName) + "." + fileType;
+    const isJpgOrPng = file.type === "image/jpeg" || file.type === "image/png";
+    if (!isJpgOrPng) {
+      message.error("You can only upload JPG or PNG images!");
```
singingwolfboy: Why this restriction?

@app/client/src/next.config.js
makoncline: Thanks for the code review! I've been programming solo so far, so its good to get feedback.  

This is the suggested way to pass server only [runtime configuration](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration) to Next.  

image  

I think..... that these are only available on the server side, since it's used only in a [next api route](https://nextjs.org/docs/api-routes/introduction).  

The intent is that this does exactly what you said, "...generate a signed upload URL on the server, and only pass that signed URL to the client".
I'll look into using the `react-s3-uploader` library.

Can you confirm if this is actually sending the serverRuntimeConfig: credentials to the client?? 
If so, I have some other things I need to change quickly 😳...
At least I'm using limited IAM roles.

@app/components/src/AvatarUpload.tsx
makoncline: Just to restrict the file type to an image format. What other image formats should be supported?

As far as I can tell, you can't restrict what type of file is uploaded with a pre signed url.  
You can give it a content type, but it will still accept any.  
Same with file size. So, thats a problem if the user want to be malicious.  
I think a different method would need to be used to enforce file size and type on the S3 end.

@app/client/src/next.config.js
singingwolfboy: Oh dear, I'm sorry for the panic. After reading the Next.js docs more carefully, I think you're right: the credentials should stay on the server.

I saw that you put the credentials in the config, and then imported and used them in a file under the `/pages` directory -- and in my limited experience with Next.js, everything under `/pages` is run on both the server _and_ the client. But I didn't know that [`serverRuntimeConfig` is special](https://nextjs.org/docs/api-reference/next.config.js/runtime-configuration) and [`/pages/api` is special](https://nextjs.org/docs/api-routes/introduction). (Seems weird to me that an API is a subset of a page...)

I tried to run the project locally, just to check for sure. Unfortunately, I ran into [a bug with graphql-codegen](https://github.com/dotansimha/graphql-code-generator/issues/3326) that prevented me from running it. However, you should be able to check yourself by [searching all of your client-side javascript in Chrome DevTools](https://stackoverflow.com/a/7970237/141395). If you search for your secret key (or even just for the string "AWSSECRETKEY", since you used that in the config), then you _should_ get no results.

@app/components/src/AvatarUpload.tsx
singingwolfboy: I don't see anything wrong with accepting all image types: that is, all MIME types that begin with `image/`. See [the MDN article on MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#Discrete_types) to see some examples.

@app/client/src/next.config.js
makoncline: No result in the search when searching for the key or any of the other server side vars.
So, at least that part seems to be secure.
I still agree that the code quality could be better.

@app/components/src/AvatarUpload.tsx
benjie: It should probably be an image format you can render in most browsers unless you're using an avatar resizing helper. You might also want to restrict things like people using GIF avatars because it can be pretty irritating...

benjie:
I tried to get this into a mergeable state in https://github.com/graphile/starter/pull/167 but I ran out of time.

PR #108, opened 2020-02-12
Extend schema to allow for presigned URLs for uploading files

singingwolfboy:
Follow-up to #107. This extends our GraphQL schema to expose an `Upload` object on the `Query` object, which allows an API client to get a pre-signed URL for uploading a file to S3. The `enabled` boolean is basically a stub right now, but you could implement it if only some users are allowed to upload files in your application. Basically, I wanted to show that there's a reason why we should have a standalone `Upload` object, instead of putting `signedUrl` directly on `Query`.


@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,37 @@
...
+import * as aws from "aws-sdk";
+import { awsRegion, uploadBucket } from "@app/config";
+
+const CreateUploadUrlPlugin = makeExtendSchemaPlugin(() => ({
+  typeDefs: gql`
+    extend type Mutation {
+      """
+      Get a signed URL for uploading files. It will expire in 60 seconds.
+      """
+      createUploadUrl(contentType: String): String!
```
benjie: Please use an input object and payload as described in [Caleb's brilliant article](https://blog.apollographql.com/designing-graphql-mutations-e09de826ed97). The payload is important for making it so you can enhance the mutation in future with additional fields; the input object in this case would be just for consistency, but it's useful to pass bundles of arguments around in a single object as the number of arguments get larger.

benjie: The exclamation points are probably the wrong way round here - content type should be required (?), but we do not guarantee a non-null result (e.g. if an error is thrown). See [why is it nullable - root fields](https://www.graphile.org/postgraphile/why-nullable/#root-query-mutation-and-subscription-fields).

benjie: Thinking about content type... Should this be an enum? We probably only want people to upload certain classes of files, rather than arbitrary files? Also the enum could be associated internally with different limits - e.g. an avatar would have small size, whereas a report or CV might have a larger file size limit. Also, they'd probably get stored to slightly different places.

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,37 @@
...
+    Mutation: {
+      async createUploadUrl(_query, args, _context, _resolveInfo) {
+        const { contentType } = args;
+        const s3 = new aws.S3({
+          region: awsRegion,
+          signatureVersion: "v4",
+        });
+        const params = {
+          Bucket: uploadBucket,
+          ContentType: contentType,
```
benjie: This should be heavily validated, I think? But if we use an enum this should be straight forward.

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,37 @@
...
+          Key: uuidv4(), // randomly generated file name
+          Expires: 60, // signed URL will expire in 60 seconds
```
benjie: 60 seconds seems short, particularly on mobile. Maybe set this to 5 minutes?

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,37 @@
...
+      async createUploadUrl(_query, args, _context, _resolveInfo) {
+        const { contentType } = args;
+        const s3 = new aws.S3({
+          region: awsRegion,
+          signatureVersion: "v4",
+        });
+        const params = {
+          Bucket: uploadBucket,
+          ContentType: contentType,
+          Key: uuidv4(), // randomly generated file name
```
benjie: Based on the enum, I think I'd add a path here. It should probably contain the users' ID.

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,37 @@
...
+          Expires: 60, // signed URL will expire in 60 seconds
+          ACL: "public-read", // uploaded file will be publicly readable
+        };
+        const signedUrl = await s3.getSignedUrlPromise("putObject", params);
```
benjie: There's no auth check, so anyone could upload any file and have it publicly accessible, and we'd be left footing the bill. Let's at least require that they're logged in. I wonder if we should just store to `/avatars/${user_id}` that way there's at most one avatar per user. We wouldn't necessarily know if it changed though...

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,37 @@
...
+          region: awsRegion,
+          signatureVersion: "v4",
+        });
+        const params = {
+          Bucket: uploadBucket,
+          ContentType: contentType,
+          Key: uuidv4(), // randomly generated file name
+          Expires: 60, // signed URL will expire in 60 seconds
+          ACL: "public-read", // uploaded file will be publicly readable
+        };
```
benjie: I think there's an option to limit file size; we should probably enable that.

benjie:
CI is failing because `aws-sdk` was not added as a dependency of the relevant package.

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,137 @@
...
+  contentType: AllowedUploadContentType;
+}
+
+const CreateUploadUrlPlugin = makeExtendSchemaPlugin(() => ({
+  typeDefs: gql`
+    """
+    The set of content types that we allow users to upload.
+    """
+    enum AllowedUploadContentType {
+      "image/apng"
```
benjie, line 30: For GraphQL documentation, this should use triple-quote syntax:

```suggestion
      """image/apng"""
```

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,137 @@
...
+      "image/tiff"
+      IMAGE_TIFF
+      "image/webp"
+      IMAGE_WEBP
+    }
+
+    """
+    All input for the \`createUploadUrl\` mutation.
+    """
+    input CreateUploadUrlInput {
```
benjie, line 51:
```suggestion
    input CreateUploadUrlInput @scope(isMutationInput: true) {
```

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,137 @@
...
+      to upload. For further information about content types, see
+      https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types
+      """
+      contentType: AllowedUploadContentType!
+    }
+
+    """
+    The output of our \`createUploadUrl\` mutation.
+    """
+    type CreateUploadUrlPayload {
```
benjie, line 68:
```suggestion
    type CreateUploadUrlPayload @scope(isMutationPayload: true) {
```

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,137 @@
...
+        context: OurGraphQLContext,
+        _resolveInfo
+      ) {
+        const { rootPgPool } = context;
+        const { input } = args;
+        const {
+          rows: [user],
+        } = await rootPgPool.query(
+          `select username from app_public.users where id = app_public.current_user_id()`
+        );
```
benjie, line 111: Since this resolver invokes SQL, the entire thing needs to be wrapped in a `SAVEPOINT` and try/catch(`ROLLBACK TO SAVEPOINT`)/finally(`RELEASE SAVEPOINT`) otherwise an error here could cause issues for other mutations in the same operation document.

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,137 @@
...
+        const contentType: string = AllowedUploadContentType[input.contentType];
+        const s3 = new aws.S3({
+          region: awsRegion,
+          signatureVersion: "v4",
+        });
+        const params = {
+          Bucket: uploadBucket,
+          ContentType: contentType,
+          // randomly generated filename, nested under username directory
+          Key: `${username}/${uuidv4()}`,
```
benjie, line 123: Let's use the users ID, which cannot change, as the key; rather than the username which can change.

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,137 @@
...
+        _resolveInfo
+      ) {
+        const { rootPgPool } = context;
+        const { input } = args;
+        const {
+          rows: [user],
+        } = await rootPgPool.query(
+          `select username from app_public.users where id = app_public.current_user_id()`
+        );
+        const username: string = user.username;
```
benjie, line 112: This would throw an ugly unhelpful error if there's no currently logged in user.

benjie:
I tried to get this into a mergeable state in https://github.com/graphile/starter/pull/167 but I ran out of time.

PR #109, opened 2020-02-13
Create dedicated URL type

singingwolfboy:
This pull request uses PostgreSQL's [domain types feature](https://www.postgresql.org/docs/current/domains.html) to define a URL type directly in the database, and modifies the `app_public.users` table to use that domain type for the `avatar_url` column. It also creates a new `graphile-engine` plugin to expose this data type in the GraphQL schema.

The benefit of this change is that GraphQL clients are informed that `avatarUrl` won't accept just _any_ string -- it must be a properly formatted URL. This was already enforced in the database, but because it wasn't exposed in the GraphQL type system, clients may not recognize this requirement until they run into a validation error.

If this pull request is accepted, then we can also use this URL type in #108.


@app/server/src/plugins/PgTypeEmailPlugin.ts
```diff
@@ -0,0 +1,149 @@
...
+        "An address in the electronic mail system. Email addresses such as `John.Smith@example.com` are made up of a local-part, followed by an `@` symbol, followed by a domain.",
+      serialize: (email: string) => email,
+      parseValue,
+      parseLiteral,
+    },
+    scope
+  );
+
+  return GraphQLEmailType;
+}
```
benjie: @singingwolfboy  It seems to me that this plugin adds the `Email` type, with a comment, and performs validation on it. However we already perform validation on the type using the domain constraint you added; so I'm not sure if these 149 lines of code add much additional value? If this plugin were removed, would the following be sufficient?

```sql
create domain app_public.email as citext check(VALUE ~ '[^@]+@[^@]+\.[^@]+');
comment on domain app_public.email is
  'An address in the electronic mail system. Email addresses such as `John.Smith@example.com` are made up of a local-part, followed by an `@` symbol, followed by a domain.';
```

benjie: I’m cool with the custom scalar and totally understand the reasoning; for some reason I was expecting creating a custom domain to result in a custom scalar being created. Seems maybe I never implemented that https://github.com/graphile/graphile-engine/blob/v4/packages/graphile-build-pg/src/plugins/PgTypesPlugin.js#L844

benjie:
I think I'd like to see this as a separate plugin, potentially a built-in option to PostGraphile, rather than adding a significant amount of code to starter to handle it. Ideally just issuing 

```sql
create domain app_public.email as citext check(VALUE ~ '[^@]+@[^@]+\.[^@]+');
comment on domain app_public.email is
  'An address in the electronic mail system. Email addresses such as `John.Smith@example.com` are made up of a local-part, followed by an `@` symbol, followed by a domain.';
```

should result in the GraphQL schema containing

```graphql
"""
An address in the electronic mail system. Email addresses such as 
`John.Smith@example.com` are made up of a local-part, followed by
an `@` symbol, followed by a domain.
"""
scalar Email
```

singingwolfboy:
That makes sense to me. Are you OK with leaving this pull request open until that feature exists in PostGraphile? That way, other people using PostGraphile are more likely to find this code and use it if they need it.

benjie:
Absolutely; I plan to leave it open :+1:

PR #113, opened 2020-02-22
Redirect if authenticated, on pages reset, register, forgot

deden:


@app/client/src/pages/forgot.tsx
```diff
@@ -7,20 +7,27 @@ import React, {
...
 import { ApolloError } from "apollo-client";
 import { getCodeFromError, extractError } from "../errors";
+import { Redirect } from "@app/components";
 
 const ForgotPassword: NextPage = () => {
   const [error, setError] = useState<Error | ApolloError | null>(null);
   return (
     <SharedLayout title="Forgot Password">
-      <WrappedForgotPasswordForm error={error} setError={setError} />
+      {({ currentUser }: SharedLayoutChildProps) =>
```
benjie: The `: SharedLayoutChildProps` cast shouldn't be required?
```suggestion
      {({ currentUser }) =>
```

deden: Yes, that's right. I've updated it

deden:
Sounds good to me. I've tried implementing it in `SharedLayout` prop with a `AuthRestrict` enum.
what do you think?

@app/client/src/layout/SharedLayout.tsx
```diff
@@ -86,6 +98,14 @@ function SharedLayout({ title, noPad = false, children }: SharedLayoutProps) {
       ) : (
         children
       );
+    if (data && data.currentUser && forbidWhen == AuthRestrict.LOGGED_IN) {
+      return (
+        <StandardWidth>
+          <Redirect href={"/"} />
+        </StandardWidth>
+      );
+    }
```
benjie: Nice! This needs to come after `useSharedLayoutQuery` of course, otherwise there's no `data` variable and branching is forbidden. What do you think about redirecting to the parent path rather than `/`? e.g. if they're on `/o/myorg/settings` we could just redirect to `/o/myorg` rather than `/`?

We also need to add something like:

```suggestion
    }
    if (data && data.currentUser === null && !loading && !error && forbidWhen === AuthRestrict.LOGGED_OUT) {
      return <Redirect href={`/login?next=${encodeURIComponent(fullHref)}`} />
    }
```

See https://github.com/graphile/starter/blob/82f5f8d5155877b3e22c65a4759f48ef5e326856/%40app/client/src/layout/SettingsLayout.tsx#L71 for an example of constructing this URL. Also, probably remove this functionality from SettingsLayout and move it wholesale to SharedLayout.

deden: yes, cool, I almost forgot with the possibility of that branch organizations-1 URL pattern

benjie: Yeah sorry, really need to get that merged!

deden: Hi @benjie, I don't think it will easy to implement right now, because this `organization.slug` does not yet exist yet in the current branch, right? :)

benjie: It shouldn't be specific to organizations; it just redirects to `?next={THE_CURRENT_URL_HERE}` - it should work for all pages existing and future.

benjie:
I've merged master and added the LOGGED_OUT branch. Merging after organizations-1 was non-trivial so I thought it fairer if I did it myself :sweat_smile:

benjie:
Reviewed all `<SharedLayout` uses and enhanced a few more behaviours... I think I co-opted this PR a little :laughing:

PR #114, opened 2020-02-22
Update README with connection to db via psql

eleijonmarck:
For newbies <-- it would be great to have a clear example of how to connect to the database and see the schema.
This is essential I would say to later than see how the impact of a change in `current.sql` will take immediate effect (i.e. migration happening in realtime).

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -51,7 +51,7 @@ run
 To connect to the newly created container db with psql/{db tool} you connect to it by default if not anything else specified via port 6543 on localhost with the `DATABASE_NAME` specificed as such:
 
 ```bash
-$ psql postgres://$DATABASE_NAME@localhost:6543
+$ psql "postgres://$DATABASE_OWNER:$DATABASE_OWNER_PASSWORD@localhost:6543/$DATABASE_NAME"
 ```
 
 #### Use Case Example:
--- a/docker/README.md
+++ b/docker/README.md
@@ -48,7 +48,7 @@ run
 
 #### DB tool:
 
-To connect to the newly created container db with psql/{db tool} you connect to it by default if not anything else specified via port 6543 on localhost with the `DATABASE_NAME` specificed as such:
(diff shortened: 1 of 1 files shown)


docker/README.md
```diff
@@ -46,6 +46,14 @@ and may make it harder to reproduce issues.
 run
 [docker-volume-watcher](https://github.com/merofeev/docker-windows-volume-watcher)
 
+#### DB tool:
+
+To connect to the newly created container db with psql/{db tool} you connect to it by default if not anything else specified via port 6543 on localhost with the `DATABASE_NAME` specificed as such:
+
+```bash
+$ psql postgres://$DATABASE_NAME@localhost:6543
```
benjie: The database name goes into the path of the connection string, and you may need to add authentication credentials. These settings are from `.env`:

```suggestion
$ psql "postgres://$DATABASE_OWNER:$DATABASE_OWNER_PASSWORD@localhost:6543/$DATABASE_NAME"
```

docker/README.md
```diff
@@ -46,6 +46,14 @@ and may make it harder to reproduce issues.
 run
 [docker-volume-watcher](https://github.com/merofeev/docker-windows-volume-watcher)
 
+#### DB tool:
+
+To connect to the newly created container db with psql/{db tool} you connect to it by default if not anything else specified via port 6543 on localhost with the `DATABASE_NAME` specificed as such:
```
benjie:
```suggestion
To connect to the container database with psql or another database tool, use port 6543 on localhost and populate the `DATABASE_NAME`, `DATABASE_OWNER` and `DATABASE_OWNER_PASSWORD` from your `.env` file:
```

docker/README.md
eleijonmarck: Cool! Do we need to say something about .env you think?

eleijonmarck:
@benjie did something happen between the fixes?

docker/README.md
benjie: Not sure

benjie:
It's a lint error, you need to run `yarn lint:fix` and commit the result :+1:

PR #117, opened 2020-02-26
Fix docker postgres issues

### LINKED ISSUE #110: yarn docker setup fails on MacOS
Setup:
- MacOS Catalaina
- yarn v1.21.1
- ZSH & iTerm

Steps to reproduce:
- git clone https://github.com/graphile/starter.git
- yarn docker setup
yarn docker setup output

I also tried pruned all containers and images to force rebuild but it did not help.
Found this issue https://github.com/graphile/starter/issues/3 and checked commit history but did not find anything useful in this date range which was very different to the current setup

generated .env file

benjie:
Fixes #110 

@JoeSchr can you cast your eyes over this? The `postgres:11` image now requires the `POSTGRES_PASSWORD` envvar to be set. I've made this work with the standard `yarn docker setup && yarn docker start` flow, but not sure if we need to do anything to make it work with devcontainers?


docker/docker.env
```diff
@@ -7,9 +7,6 @@ FORCE_COLOR=2
 # `.gmrc` checks for interactive TTY.
 PG_DUMP=docker-compose exec -T db pg_dump
 
-# Instructs to use our docker PostgreSQL service.
-DATABASE_HOST=db
-ROOT_DATABASE_URL=postgres://postgres@db/template1
```
JoeSchr: I don't think we should remove that, because it's there to override the .env vars when using docker-compose

docker/scripts/yarn-setup.js
```diff
@@ -1,6 +1,10 @@
 const { spawnSync: baseSpawnSync } = require("child_process");
 const { basename, dirname, resolve } = require("path");
 const platform = require("os").platform();
+const { safeRandomString } = require("../../scripts/lib/random");
+const fsp = require("fs").promises;
+
+const DOTENV_PATH = `${__dirname}/../../.env`;
```
JoeSchr: or we should just use `docker.env` here instead of `.env`

why does it need to be in `.env`? it's only relevant for docker right?

benjie: Basically: `docker.env` is committed, `.env` is not, and we need `ROOT_DATABASE_URL` to contain the password we've generated for `POSTGRES_PASSWORD` so we cannot write it to `docker.env` otherwise it'd be tracked under git. So instead I just have Postgres source the `.env` file.

Oh... I see what you mean - this breaks dual Docker/local dev. I don't think that's an issue, TBH - we've told users to pick one and stick to it in the README so it should be fine? Maybe we should add a `docker/.env` file that contains these secrets and source both, though. It looks like (though the docs don't say) `env_file` could take an array. I'm going to give this a try.

benjie:
@JoeSchr based on your feedback, I've removed `docker/docker.env` and replaced it with `docker/.env` which is a secret file (untracked) that the `yarn docker setup` command populates automatically. I did consider keeping both a tracked and untracked .env but I decided that was too much.

JoeSchr:
I'm kind of using dual docker/local mode all the time nowadays, so I would
be glad if it didn't break. but if it's needed I will work around it


__

Typed on a mobile, please excuse any typing mistakes or short-spokeness.

benjie:
I think the above solves this :+1: Let me know if it doesn't.

JoeSchr:
I think so too, just wanted to stated it as an use case that's actively used

PR #121, opened 2020-03-09
Add indexes to posts so that they resolve as expected

bdombro:
As a newcomer to Postgraphile, I was confused that posts didn't resolve relations despite the documentation suggesting it. This PR adds the required indexes, so that future users will (hopefully) be less confused.

### DIFF
--- a/@app/db/migrations/current.sql
+++ b/@app/db/migrations/current.sql
@@ -108,8 +108,8 @@ relevant operations on them. The tables will appear when you uncomment the
 --   created_at       timestamptz not null default now()
 -- );
 -- alter table app_public.user_feed_posts enable row level security;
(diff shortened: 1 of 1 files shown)


@app/db/migrations/current.sql
```diff
@@ -107,6 +108,8 @@ relevant operations on them. The tables will appear when you uncomment the
 --   created_at       timestamptz not null default now()
 -- );
 -- alter table app_public.user_feed_posts enable row level security;
+-- create index on app_public.user_feed_posts (user_id); -- Add index to post so that user can be resolved
+-- create index on app_public.user_feed_posts (post_id); -- Add index to post so that post can be resolved
```
benjie:
```suggestion
-- create index on app_public.user_feed_posts (user_id);
-- create index on app_public.user_feed_posts (post_id);
```

PR #127, opened 2020-03-20
Organizations

benjie:
This pull request adds a significant feature to Graphile Starter: Organizations. Organizations are useful for the vast majority of B2B SaaS projects as it allows multiple people to securely access the same resources, each potentially with their own roles. We've only introduced the "owner" (i.e. admin) and "billing contact" roles (each of which must be assigned to exactly one user), but you can of course extend the schema to add your own. The PR handles creating organizations, switching between organizations, deleting organizations, inviting people to organizations and much more. Another thing this PR does is change the database to use UUIDs for most primary keys to help mask how many user accounts your system has (this has been a commonly requested feature).


benjie:
I think this is pretty much ready to merge, feature-wise.

PR #130, opened 2020-03-23
test(scripts): exit the test script with the same code as jest output

charlex:


benjie:
Did you close this deliberately?

charlex:
@benjie yes, I meant to open my [own internal PR](https://github.com/charlex/starter/pull/1).

I was going to talk to you (or open an issue) to see if you thought it was an actual issue or not.

I tend to think this is an issue, and that `starter`'s CI is probably returning false positives. But I could just be missing something simple..

charlex:
Here is an example of a false positive I triggered: https://github.com/charlex/starter/runs/528093596?check_suite_focus=true

charlex:
Guess I'll open this since we're talking about it now anyways..

benjie:
Yeah; this is a bug. I thought `spawnSync` threw on non-zero but it doesn't.

charlex:
Yeah, I thought the same thing. It confused me very much to discover what was happening!

PR #146, opened 2020-04-26
WIP: fix(@app/server): do not prepare next app on each request

charlex:
Filing tentatively... I need to check the logging to ensure that next is in fact being prepared on each request.


benjie:
It’s not being initialised on each request: a single promise is made, and the middleware awaits that promise. This is so the rest of the system can get up and running (e.g. start serving the graphql endpoint) whilst Next sorts itself out; any requests for next stuff will hang until the promise resolves. After that, awaiting a resolved promise is a fairly efficient option, I decided it wasn’t worth complicating the code to save a microtask per request.

PR #167, opened 2020-05-15
Upload avatar

benjie:
This is a work in progress PR for adding avatar upload functionality, merging #101 and #108. It's no-where near ready.

- [x] Add createSignedUrl mutation 
- [x] Add upload component
- [x] Render profile page in two seconds - one for avatar, one for textual settings
- [x] Handle errors during avatar upload flow
- [ ] Document how to configure IAM, S3, etc for uploads
- [ ] Allow configuring the system to not allow avatar uploads
- [ ] Uploads should be private and expire automatically after a few hours
- [ ] The uploaded URL should be sent to the server, validated, and then the object should be moved to public S3 that doesn't expire, and referenced in the DB so that it may later be safely deleted (beware of race conditions)
- [ ] A trigger should be added such that when a new upload is performed, the old upload gets queued for deletion
- [ ] Storage needs to be added to the database to track the uploads the user has permissions to delete

Unfortunately I timed out after 4 hours merging these PRs and attempting to implement the above functionality.

### DIFF
--- a/@app/server/src/plugins/CreateUploadUrlPlugin.ts
+++ b/@app/server/src/plugins/CreateUploadUrlPlugin.ts
@@ -50,47 +50,20 @@ async function getCurrentUser(pool: Pool): Promise<User | null> {
     await pool.query("RELEASE SAVEPOINT");
   }
 }
+/** The set of content types that we allow users to upload.*/
+const ALLOWED_UPLOAD_CONTENT_TYPE = [
+  "image/apng",
+  "image/bmp",
+  "image/gif",
+  "image/jpeg",
+  "image/png",
+  "image/svg+xml",
+  "image/tiff",
+  "image/webp",
+];
 
 const CreateUploadUrlPlugin = makeExtendSchemaPlugin(() => ({
   typeDefs: gql`
-    """
-    The set of content types that we allow users to upload.
-    """
-    enum AllowedUploadContentType {
-      """
-      image/apng
-      """
-      IMAGE_APNG
-      """
-      image/bmp
-      """
-      IMAGE_BMP
-      """
-      image/gif
-      """
-      IMAGE_GIF
-      """
-      image/jpeg
-      """
-      IMAGE_JPEG
-      """
-      image/png
-      """
-      IMAGE_PNG
-      """
-      image/svg+xml
-      """
-      IMAGE_SVG_XML
-      """
-      image/tiff
-      """
-      IMAGE_TIFF
-      """
-      image/webp
-      """
-      IMAGE_WEBP
-    }
-
     """
     All input for the \`createUploadUrl\` mutation.
     """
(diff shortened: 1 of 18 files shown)


jackbravo:
I tried running with a [minio](url) server, which you can run with something like:

```
mkdir /tmp/s3
docker run -p 9000:9000 --name minio1 \
  -v /tmp/s3:/data \
  minio/minio server /data
```

Then adding `S3_UPLOADS_BUCKET=http://localhost:9000/uploads` to my .env file.

But then, while trying to upload a file I get this error in the javascript console:

```
[GraphQL error]: message: syntax error at end of input, location: [{"line":2,"column":3}], path: ["createUploadUrl"]
Error: "GraphQL error: syntax error at end of input"
  code: "42601"
​  severity: "ERROR"
```

lukeramsden:
I got the same error as the previous commenter - I believe it is something to do with the `SAVEPOINT` in `getCurrentUser` not having a name, because changing the 3 queries to be `[..] SAVEPOINT create_upload_url` gives me a different error:

```
{
  "errors": [
    {
      "message": "SAVEPOINT can only be used in transaction blocks",
      "locations": [
        {
          "line": 25,
          "column": 3
        }
      ],
      "path": [
        "createUploadUrl"
      ],
      "extensions": {
        "exception": {
          "code": "25P01",
          "severity": "ERROR"
        }
      }
    }
  ],
  "data": {
    "createUploadUrl": null
  }
}
```

So when we call `getCurrentUser` it's not inside a transaction? That's weird because the docs say[0] that all mutations should run under a transaction. Could this be caused by using the wrong version of Postgraphile or Postgres?

[0] https://www.graphile.org/postgraphile/make-extend-schema-plugin/#querying-the-database-inside-a-resolver

@app/server/src/plugins/CreateUploadUrlPlugin.ts
```diff
@@ -0,0 +1,170 @@
...
+        if (!uploadBucket) {
+          const err = new Error(
+            "Server misconfigured: missing `S3_UPLOADS_BUCKET` envvar"
+          );
+          // @ts-ignore
+          err.code = "MSCFG";
+          throw err;
+        }
+
+        const user = await getCurrentUser(context.rootPgPool);
```
lukeramsden, line 126: I believe this is what is causing the input error issue, changing `rootPgPool` to `pgClient` fixes it.

benjie:
Yes, the savepoint should have a name. The whole `getCurrentUser` function doesn't currently make sense; e.g. there's no point doing `pool.query('savepoint...')` because the client you `pool.query('release savepoint...')` from may not be the same client that created the savepoint.

Basically: this PR is a very WIP merge of two other PRs and is not suitable for usage - that's why it's in draft status.

PR #174, opened 2020-06-14
Converted SharedLayout->AuthRestrict to flag enum to allow for multiple forbid conditions

lukeramsden:
Also added IS_ADMIN and NOT_ADMIN flags.
This is better because it allows for multiple forbid conditions, as well as easy extension.
The IS_ADMIN/NOT_ADMIN check is kinda ugly but that's how it's auto-formatted so I've kept it like that for this PR.


lukeramsden:
Good points, I've made some changes.

PR #192, opened 2020-08-11
Modify components to support ionic fork

justinr1234:


@app/lib/src/withApollo.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+const withApolloWithoutNext = (Component: any, options?: WithApolloOptions) => (
+  props: any
+) => {
+  const apollo = getApolloClient({}, options);
+  return <Component {...props} apollo={apollo} />;
+};
+
+export const withApollo = (Component: any, options?: WithApolloOptions) =>
+  options?.useNext === false
```
justinr1234, line 129: Optionally disable nextjs processing for apollo, but enable it by default

@app/server/src/middleware/installCSRFProtection.ts
```diff
@@ -1,5 +1,20 @@
 import csrf from "csurf";
 import { Express } from "express";
+import url from "url";
+
+const skipList = process.env.CSRF_SKIP_REFERERS
```
justinr1234, line 5: Allow us to pass a list of referrers like:

```
CSRF_SKIP_REFERERS=localhost:8100,mobile.mydomain.com
```

@app/server/src/middleware/installCors.ts
```diff
@@ -0,0 +1,19 @@
+import cors from "cors";
+import { Express } from "express";
+
+export default (app: Express) => {
+  const origin = [];
+  if (process.env.ROOT_URL) {
+    origin.push(process.env.ROOT_URL);
+  }
+  if (process.env.CORS_ALLOWED_URLS) {
```
justinr1234, line 9: Allow us to pass a list of domains allowed for CORS:

```
CORS_ALLOWED_URLS=http://localhost:8100,https://mobile.mydomain.com
```

@app/server/src/app.ts
```diff
@@ -99,6 +99,7 @@ export async function makeApp({
    * express middleware. These helpers may be asynchronous, but they should
    * operate very rapidly to enable quick as possible server startup.
    */
+  await middleware.installCors(app);
```
justinr1234, line 102: Require CORS to allow ionic dev site to communicate with our graphql API

@app/lib/src/withApollo.tsx
```diff
@@ -0,0 +1,131 @@
...
+    httpLink
+  );
+  return mainLink;
+}
+
+const getApolloClient = (
+  { initialState, ctx }: InitApolloOptions<NormalizedCacheObject>,
+  withApolloOptions?: WithApolloOptions
+): ApolloClient<NormalizedCacheObject> => {
+  const ROOT_URL = process.env.ROOT_URL || withApolloOptions?.rootUrl;
```
justinr1234, line 78: Allow us to pass in a different rootUrl in the options to `withApollo`

@app/lib/src/withApollo.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { WebSocketLink } from "apollo-link-ws";
+import { getOperationAST } from "graphql";
+import withApolloBase, { InitApolloOptions } from "next-with-apollo";
+import React from "react";
+import { SubscriptionClient } from "subscriptions-transport-ws";
+import ws from "ws";
+
+import { GraphileApolloLink } from "./GraphileApolloLink";
+
+interface WithApolloOptions {
```
justinr1234, line 16: `withApollo` can now be called as:

```
withApollo(App, {
  useNext: false,
  rootUrl: process.env.REACT_APP_ROOT_URL,
});
```

.eslintignore
```diff
@@ -4,3 +4,4 @@ node_modules
 /data/schema.sql
 /@app/graphql/index.*
 /@app/client/.next
+**/build
```
justinr1234, line 7: Ignore any build folders, such as the one ionic generates

package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "description": "Description of project here",
   "scripts": {
-    "setup": "yarn && node ./scripts/setup.js",
+    "setup": "yarn && node ./scripts/setup.js && lerna run setup",
```
justinr1234, line 7: Allow other packages to run setup tasks, such as the mobile package so it can add custom env vars

@app/server/src/middleware/installSession.ts
```diff
@@ -75,7 +75,13 @@ export default (app: Express) => {
    * different authentication method such as bearer tokens.
    */
   const wrappedSessionMiddleware: RequestHandler = (req, res, next) => {
-    if (req.isSameOrigin) {
+    const origins = [];
+    if (process.env.SESSION_ALLOWED_ORIGINS) {
```
justinr1234, line 79: Allow session cookies to be sent from origins other than the main client. Give the list of allowed origins.

benjie:
As it stands, this PR contains a lot of changes; as discussed on Discord, lets break it up into a number of bite-sized PRs that will be easier to review and iterate so we can merge things a bit faster. Please make sure that each PR comes with an explanation why this change is valuable in general (not specifically limited to Ionic, e.g. you might show how the change is also necessary for Create React App). Here's some PR topics that would make sense from this:

### Formatting changes

I notice that `cypress.yml` had the `option` re-wrapped; this *should* be handled automatically by `prettier` which formats all our yaml for us, but if this is not the case, a change that enforces prettier formatting of the relevant files would be welcome.

### sudo apt-get update

This fix should be in place for everyone, please send it as a standalone PR.

### withApollo.ts -> .tsx

A simple rename of this file could be included such that the diff when you later edit it will be much clearer. Actually; since this is just a file rename I'll do it now... Done.

### cors

This is going to be a controversial addition which will require quite a bit of back-and-forth (but I'm generally in favour of merging, when done right); lets get it as a clean PR on its own.

### lerna run setup

Raise this as a separate PR including a small mention in the README. If it passes tests I see no reason not to merge it.

### Session changes (allowed origins)

Please raise a PR for just this change with a good explanation of why it's needed, whether it's needed only for development, and any security concerns/caveats there may be with it's introduction.

justinr1234:
@benjie updated

benjie:
I missed the CSRF stuff; we should split that out too.

PR #195, opened 2020-09-02
Add CORS with allowed origins

justinr1234:
This adds CORS to the app. It allows you to specify, via environmental variables, which allowed origins you can use. This is useful for a variety of scenarios and is specifically useful in the case of using this server for an API or accessing it from other origins (like using Ionic).

Allow us to pass a list of domains allowed for CORS:

```
CORS_ALLOWED_URLS=http://localhost:8100,https://mobile.mydomain.com
```


@app/server/src/middleware/installCors.ts
```diff
@@ -0,0 +1,19 @@
...
+  if (process.env.CORS_ALLOWED_URLS) {
+    origin.push(
+      ...(process.env.CORS_ALLOWED_URLS?.replace(/s\s/g, "").split(",") || [])
+    );
+  }
+  const corsOptions = {
+    origin,
+    credentials: true,
+  };
+  app.use(cors(corsOptions));
```
benjie, line 18: I'm not sure your regexp `/s\s/g` is correct... Why is the literal `s` in there? Also I think we could compress this down a bit, improving the TypeScript inference/error messages without making it harder to read:

```suggestion
  app.use(cors({
    origin: [
      ...(process.env.ROOT_URL ? [process.env.ROOT_URL] : []),
      ...(process.env.CORS_ALLOWED_URLS?.replace(/\s+/g, "").split(",") || [])
    ],
    credentials: true,
  }}));
```

benjie (changes requested):
I'd like to gate CORS on the envvar, so by default CORS is not enabled.

benjie:
_[semi-automated message]_ Hi, there has been no activity in this issue for a while so I'm closing it to keep the issues/pull requests manageable. If this is still an issue, please re-open with additional details.

PR #199, opened 2020-09-05
fix: user id should be type `uuid` not `int`

JofArnold:
In the commented sample code `user_feed_posts.user_id` and `posts.author_id` is defined as type `int`. However `current_user_id` returns type `uuid` so these should also be `uuid`.

I can confirm the example works with this fix when fully uncommented


benjie (approved):
Awesome; thanks for spotting this! 🙌 

(We used to use `serial` but switched to `uuid` a while back, clearly this didn't get updated.)

PR #202, opened 2020-09-21
fix(types): graphql-codegen avoid optionals

benjie:


acao (approved):

heri16:
Why is this required?

benjie:
Gosh; spam much? There’s really no need to ask this same question in three different places, it’s extremely rude.

It improves the accuracy of the types.

PR #203, opened 2020-09-22
feat: Windows Support

MegaCookie:
Hi there! It's a cool project, however I had some issues running this on Windows.

I added support for setting environmnent variables on Windows, and using [npx](https://www.npmjs.com/package/npx) for npm package binaries.   

Next I added NodeJS scripts that can be ran by node for `.gmrc` of _graphile_migrate_ so it doesn't rely on Bash support, and made some improvements for making sure all commands can be ran correctly on Windows.


@app/client/package.json
```diff
@@ -3,8 +3,8 @@
   "private": true,
   "version": "0.0.0",
   "scripts": {
-    "build": "cd src && node -r @app/config/env ../node_modules/.bin/next build",
-    "test": "NODE_ENV=test node -r @app/config/env ./node_modules/.bin/jest"
+    "build": "cd src && npx -n \"-r @app/config/env\" next build",
+    "test": "cross-env NODE_ENV=test npx -n \"-r @app/config/env\" jest"
```
benjie, line 7: When using npx we don't want to risk it doing temporary installs of different versions just because the user's node_modules cache is out of date (or, for example, they're using yarn2 and the files don't exist where npx looks for them). Let's add the `--no-install` flag:

```suggestion
    "build": "cd src && npx --no-install -n \"-r @app/config/env\" next build",
    "test": "cross-env NODE_ENV=test npx --no-install -n \"-r @app/config/env\" jest"
```

@app/db/.gmrc
```diff
@@ -68,7 +68,7 @@
     "!afterReset.sql",
     {
       "_": "command",
-      "command": "DATABASE_URL=\"${GM_DBURL}\" yarn workspace @app/worker install-db-schema"
+      "command": "node scripts/after-reset.js"
```
benjie, line 71: I'm not keen to add a JS script that then just forks out to another command line; it's a bit of a pain to maintain and the indirection makes it harder for the user to follow. Is it possible to accomplish this with `cross-env`?

```suggestion
      "command": "cross-env-shell \"DATABASE_URL=\\\"$GM_DBURL\\\" yarn workspace @app/worker install-db-schema\""
```

@app/db/.gmrc
```diff
@@ -79,7 +79,7 @@
     {
       "_": "command",
       "shadow": true,
-      "command": "if [ \"$IN_TESTS\" != \"1\" ]; then ./scripts/dump-db; fi"
+      "command": "node scripts/dump-db.js"
```
benjie, line 82: This makes the user think we always dump db; let's add a comment to clarify:

```suggestion
      // NOTE: this script does nothing when envvar `IN_TESTS` is `1`
      "command": "node scripts/dump-db.js"
```

@app/db/.gmrc
```diff
@@ -91,7 +91,7 @@
     {
       "_": "command",
       "shadow": true,
-      "command": "if [ \"$IN_TESTS\" = \"1\" ]; then node scripts/test-seed.js; fi"
+      "command": "node scripts/test-seed.js"
```
benjie, line 94:
```suggestion
      // NOTE: this script does nothing unless envvar `IN_TESTS` is `1`
      "command": "node scripts/test-seed.js"
```

@app/client/package.json
benjie: (Please apply this change throughout the entire PR)

@app/db/package.json
```diff
@@ -3,21 +3,22 @@
...
+    "gm": "npx -n \"-r @app/config/env\" graphile-migrate",
     "migrate": "yarn gm migrate",
     "watch": "yarn gm watch",
     "commit": "yarn gm commit",
     "uncommit": "yarn gm uncommit",
     "reset": "yarn gm reset",
     "dump": "yarn gm migrate && yarn gm reset --shadow --erase && yarn gm migrate --shadow --forceActions",
     "wipe-if-demo": "./scripts/wipe-if-demo",
-    "test": "NODE_ENV=test node -r @app/config/env ./node_modules/.bin/jest"
+    "test": "cross-env NODE_ENV=test npx -n \"-r @app/config/env\" jest"
```
benjie, line 14: I wish there was a more efficient way to support Windows; at the moment this is spawning a Node process `cross-env` which spawns a Node process `npx` which then spawns _another_ Node process `jest`. So we've got a stack of 3 node processes, where previously we only had one. Is there a particular reason we need to use `npx` here for Windows support?

@app/db/scripts/after-reset.js
```diff
@@ -0,0 +1,9 @@
+const { spawn } = require("child_process");
+spawn("yarn", ["workspace", "@app/worker", "install-db-schema"], {
+  stdio: "inherit",
+  env: {
+    ...process.env,
+    DATABASE_URL: process.env.GM_DBURL,
+  },
+  shell: true,
+});
```
benjie, line 9: Hopefully we can replace this script with a command-line (see comment above)

@app/db/scripts/dump-db.js
```diff
@@ -0,0 +1,31 @@
...
+    ],
+    {
+      stdio: "inherit",
+      env: {
+        ...process.env,
+      },
+      shell: true,
+    }
+  );
+}
```
benjie, line 31: Please delete the old `dump-db` script.

@app/db/scripts/dump-db.js
```diff
@@ -0,0 +1,31 @@
+const { spawn } = require("child_process");
+
+const inTests = process.env.IN_TESTS;
+if (!inTests) {
+  const connectionString = process.env.GM_DBURL;
+  if (!connectionString) {
+    throw new Error(
+      "This script should only be ran from inside graphile-migrate"
+    );
```
benjie, line 9: No need for a stack trace here.

```suggestion
    console.error(
      "This script should only be called from a graphile-migrate action."
    );
    process.exit(1);
```

@app/db/scripts/dump-db.js
```diff
@@ -0,0 +1,31 @@
...
+      "--exclude-schema=graphile_migrate",
+      "--exclude-schema=graphile_worker",
+      "--file=../../data/schema.sql",
+      `"${connectionString}"`,
+    ],
+    {
+      stdio: "inherit",
+      env: {
+        ...process.env,
+      },
```
benjie, line 27: No need for this, the default is `process.env` and here we're cloning that object for no benefit.
```suggestion
```

@app/db/scripts/dump-db.js
```diff
@@ -0,0 +1,31 @@
+const { spawn } = require("child_process");
+
+const inTests = process.env.IN_TESTS;
+if (!inTests) {
```
benjie, line 4: Let's exit early instead, so we don't have to look for an `else` block at the end of the file when reading the code.

```suggestion
if (process.env.IN_TESTS === "1") {
  process.exit(0);
}
```

@app/db/scripts/test-seed.js
```diff
@@ -18,7 +18,10 @@ async function main() {
   }
 }
 
-main().catch((e) => {
-  console.error(e);
-  process.exit(1);
-});
+const inTests = process.env.IN_TESTS;
+if (inTests) {
```
benjie, line 22: Please move this block either to the top of the file, or to the top of the `main` function body.

package.json
```diff
@@ -12,12 +12,14 @@
...
     "test": "node scripts/test.js",
     "posttest": "lerna run posttest",
-    "test:watch": "node scripts/test.js --watch",
+    "test:watch": "run-script-os",
+    "test:watch:darwin:linux": "sleep 10 && yarn test:watch:default",
+    "test:watch:default": "node scripts/test.js --watch",
     "lint": "tsc -b && yarn prettier:check && eslint --ext .js,.jsx,.ts,.tsx,.graphql .",
-    "lint:fix": "eslint --ext .js,.jsx,.ts,.tsx,.graphql . --fix; prettier --ignore-path .eslintignore --write '**/*.{js,jsx,ts,tsx,graphql,md}'",
-    "prettier:check": "prettier --ignore-path .eslintignore --check '**/*.{js,jsx,ts,tsx,graphql,md}'",
+    "lint:fix": "eslint --ext .js,.jsx,.ts,.tsx,.graphql . --fix && prettier --ignore-path .eslintignore --write \"**/*.{js,jsx,ts,tsx,graphql,md}\"",
```
benjie, line 19: If we're going with an `&&` here, we should change the order so that prettier always runs (as it does currently) since it's rare for prettier to fail and I definitely want `lint:fix` to format my code even if there's ESLint errors. I have concerns that some ESLint rules might attempt formatting which prettier would undo, so if this _does_ cause reformatting (it shouldn't) we should track down the rules causing this and disable them. Or we could `prettier && eslint && prettier` but that seems inefficient.

package.json
```diff
@@ -12,12 +12,14 @@
...
+    "prettier:check": "prettier --ignore-path .eslintignore --check \"**/*.{js,jsx,ts,tsx,graphql,md}\"",
     "depcheck": "lerna exec --stream 'yarn depcheck --ignores=\"@app/config,@app/client,tslib,webpack,babel-plugin-import,source-map-support,@graphql-codegen/*,graphql-toolkit,net,tls,dayjs\" --ignore-dirs=\".next\"'",
-    "dev": "yarn && lerna run codegen --stream && tsc -b && concurrently --kill-others --names 'TSC,WATCH,RUN,TEST' --prefix '({name})' --prefix-colors 'yellow.bold,yellow.bold,cyan.bold,greenBright.bold' 'tsc -b --watch --preserveWatchOutput --assumeChangesOnlyAffectDirectDependencies' 'lerna run --parallel watch' 'lerna run --parallel dev' 'sleep 10; yarn test:watch'",
+    "dev": "yarn && lerna run codegen --stream && tsc -b && concurrently -k -n \"TSC,WATCH,RUN,TEST\" -p \"({name})\" -c \"yellow.bold,yellow.bold,cyan.bold,greenBright.bold\" \"tsc -b --watch --preserveWatchOutput\" \"lerna run --parallel watch\" \"lerna run --parallel dev\" \"yarn test:watch\"",
```
benjie, line 22: Please use expanded option flags, it makes it clearer to people what the flags do when they're not familiar with the tool.

What's the reason we can't use single quotes here; does Windows not support them? Bit frustrating to have to backslash escape everything.

benjie (changes requested):
Really great that someone's finally working on proper Windows support; previously people have just used Docker on Windows I think! Thanks for working on this :raised_hands: 

I think some of these changes might not be necessary for Windows support; if so it'd be nice to trim out the ones that are more personal preference (e.g. using `npx` rather than calling `./node_modules/.bin/`, using short flags instead of long flags) so that the PR only does the changes that are absolutely necessary for Windows support. The other changes could be evaluated via a separate PR.

benjie:
Also, critically, we need some tests for this so that we know it will continue to work as Starter functionality expands and grows. Can you add some GitHub Actions tests on Windows platform? You don't _have_ to do this, but without it Windows support may reduce over time.

@app/db/package.json
MegaCookie: Unfortunately, I don't know a better way to handle these test cases on Windows, as we need to set environment variable, and only using `npx` can find the proper `jest` command on Windows.

package.json
MegaCookie: I changed it to `&&` because `;` is a single `&` on Windows. Therefore I replaced it with `run-script-os` for changing the command to a Windows specific version.

package.json
MegaCookie: Yes, I agree it is a bit cumbersome to add escaping, unfortunately it is the only way Windows supports those commands right now.

MegaCookie:
> 
> 
> Also, critically, we need some tests for this so that we know it will continue to work as Starter functionality expands and grows. Can you add some GitHub Actions tests on Windows platform? You don't _have_ to do this, but without it Windows support may reduce over time.

That is a good suggestion! I added a new workflow which uses just Node 14 and the default installed PostgreSQL from a GitHub hosted runner. I think just testing the basic support for one Node version and the default PostgreSQL version from `windows-latest` should work fine for telling if Windows support is still working.

In the end it is quite some work to get this all working with Windows in a clean way. In the end suggesting to use WSL might be also an alternative.

hiddenboox:
@MegaCookie please review one some fix in your fork https://github.com/MegaCookie/starter/pull/2

MegaCookie:
Thanks @hiddenboox for pointing me in the right direction that the Graphile Migrate afterReset command was not working right. In the end just a normal `cross-env` call in `.gmrc` did work and solved the issue with the migration.

.github/workflows/windows-nodejs.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Start PostgreSQL
+        run: |
+          $postgreSqlSvc = Get-Service "postgresql*"
+          Set-Service $postgreSqlSvc.Name -StartupType manual
+          $postgreSqlSvc.Start()
+          $env:PATH += ";$env:PGBIN"
+          $env:PGPASSWORD = "root"
+          $postgreSqlSvc.WaitForStatus('Running')
+          psql -U postgres -d postgres -c "alter user postgres with password 'postgres'";
+          Restart-Service $postgreSqlSvc
```
benjie, line 31: Is there a way to pin this to a specific PostgreSQL version? I'm concerned that jumping up major versions without explicitly stating to do so may make maintainance of the repo more challenging.

@app/db/.gmrc
```diff
@@ -68,7 +68,7 @@
     "!afterReset.sql",
     {
       "_": "command",
-      "command": "DATABASE_URL=\"${GM_DBURL}\" yarn workspace @app/worker install-db-schema"
+      "command": "cross-env DATABASE_URL=$GM_DBURL yarn workspace @app/worker install-db-schema"
```
benjie, line 71: In bash, `$GM_DBURL` should be quoted to ensure that it doesn't get expanded to multiple arguments/etc. Is this problematic for cross-env?

@app/db/scripts/dump-db.js
```diff
@@ -0,0 +1,30 @@
...
+spawn(
+  process.env.PG_DUMP || "pg_dump",
+  [
+    "--no-sync",
+    "--schema-only",
+    "--no-owner",
+    "--exclude-schema=graphile_migrate",
+    "--exclude-schema=graphile_worker",
+    "--file=../../data/schema.sql",
+    `"${connectionString}"`,
```
benjie, line 24: I don't think the `"` quote marks are needed here; since `spawn` escapes all the arguments automatically. In fact, I think this will break the script on bash (not sure about Windows) because the quote marks will be escaped and included as part of the string. Should be able to just pass `connectionString` directly:

```suggestion
    connectionString,
```

package.json
```diff
@@ -12,12 +12,18 @@
...
     "test": "node scripts/test.js",
     "posttest": "lerna run posttest",
-    "test:watch": "node scripts/test.js --watch",
-    "lint": "tsc -b && yarn prettier:check && eslint --ext .js,.jsx,.ts,.tsx,.graphql .",
-    "lint:fix": "eslint --ext .js,.jsx,.ts,.tsx,.graphql . --fix; prettier --ignore-path .eslintignore --write '**/*.{js,jsx,ts,tsx,graphql,md}'",
-    "prettier:check": "prettier --ignore-path .eslintignore --check '**/*.{js,jsx,ts,tsx,graphql,md}'",
-    "depcheck": "lerna exec --stream 'yarn depcheck --ignores=\"@app/config,@app/client,tslib,webpack,babel-plugin-import,source-map-support,@graphql-codegen/*,graphql-toolkit,net,tls,dayjs\" --ignore-dirs=\".next\"'",
-    "dev": "yarn && lerna run codegen --stream && tsc -b && concurrently --kill-others --names 'TSC,WATCH,RUN,TEST' --prefix '({name})' --prefix-colors 'yellow.bold,yellow.bold,cyan.bold,greenBright.bold' 'tsc -b --watch --preserveWatchOutput --assumeChangesOnlyAffectDirectDependencies' 'lerna run --parallel watch' 'lerna run --parallel dev' 'sleep 10; yarn test:watch'",
+    "test:watch": "run-script-os",
+    "test:watch:darwin:linux": "sleep 10 && yarn test:watch:default",
```
benjie, line 16: I don't like the sleep being here; it should be in the calling location as if you want to just run `test:watch` locally (without `yarn start` running) you should be able to do so without incurring a 10 second delay. Is it necessary for it to have moved, or were you just tidying up?

benjie: The delay is just so that the startup of the `dev` script doesn't get overwhelmed due to too much going on, it can get pretty intense for a few seconds. Maybe it would be better to pass an argument, e.g. `--delay 10`, to `scripts/test.js` from `concurrently` directly, and have the test script wait 10 seconds before starting. I'm assuming this has been done because Windows doesn't have a `sleep` command?

package.json
```diff
@@ -12,12 +12,18 @@
...
+    "test:watch:default": "node scripts/test.js --watch",
+    "lint": "tsc -b && yarn prettier:check && yarn eslint:check",
+    "lint:fix": "run-script-os",
+    "lint:fix:windows": "yarn eslint:check --fix & yarn prettier:fix",
+    "lint:fix:default": "yarn eslint:check --fix; yarn prettier:fix",
+    "eslint:check": "eslint --ext .js,.jsx,.ts,.tsx,.graphql .",
+    "prettier:check": "prettier --ignore-path .eslintignore --check \"**/*.{js,jsx,ts,tsx,graphql,md}\"",
+    "prettier:fix": "prettier --ignore-path .eslintignore --write \"**/*.{js,jsx,ts,tsx,graphql,md}\"",
+    "depcheck": "lerna exec --stream \"yarn depcheck --ignores=\"@app/config,@app/client,tslib,webpack,babel-plugin-import,source-map-support,@graphql-codegen/*,graphql-toolkit,net,tls,dayjs\" --ignore-dirs=\".next\"\"",
+    "dev": "yarn && lerna run codegen --stream && tsc -b && concurrently --kill-others --names \"TSC,WATCH,RUN,TEST\" -p \"({name})\" -c \"yellow.bold,yellow.bold,cyan.bold,greenBright.bold\" \"tsc -b --watch --preserveWatchOutput\" \"lerna run --parallel watch\" \"lerna run --parallel dev\" \"yarn test:watch\"",
```
benjie, line 26: Do we need to use the short aliases for Windows, or can we stick with the longer ones that are more legible/obvious?

benjie (commented):
This is excellent @MegaCookie! Thanks for taking this on, it'll be really great to support Windows better, for sure! I really appreciate that you've minimised the changes now; I have a few questions below, some may warrant changes, but some of them maybe just an explanation since I'm not familiar with development on Windows and want to understand the reasons for the changes.

Thanks again, this feels like it's getting close to being mergable. Stellar effort :raised_hands:

.github/workflows/windows-nodejs.yml
MegaCookie: That is a fair comment! I will update it with a specific version that can be downloaded from the Windows Package Manager.

@app/db/.gmrc
MegaCookie: Yes it does not work otherwise. I am not sure if it can lead to unwanted behavior?

package.json
MegaCookie: Yes it was moved because Windows doesn't have a `sleep` command.  An JS timer would work as well, and is a bit cleaner from the `package.json` perspective.

package.json
MegaCookie: I am not entirely sure what you mean here. I split up the prettier fixes as Windows does not support running two commands using ';', so we have another `lint:fix` command that changes the syntax based on the OS. 

Or is that not what you mean?

@app/db/.gmrc
benjie: It's an issue if the envvar contains special characters; e.g. if it contains spaces (e.g. in the database password) then it will break the command. This is solveable by being careful to URL encode the GM_DBURL but it's risky to rely on this.

The [cross-env README shows this working with quotes](https://www.npmjs.com/package/cross-env#usage)*; I think it might have been the braces causing issues, try it with just quotes:

```suggestion
      "command": "cross-env DATABASE_URL=\"$GM_DBURL\" yarn workspace @app/worker install-db-schema"
```

\* well; it shows `$` placeholders working within string args, and it shows quoted envvar values, I'm hoping that these two techniques can be combined.

package.json
benjie: Oh sorry! You changed `concurrently` to use `-p` and `-c` rather than `--prefix` and `--prefix-colors`. I prefer long flags because they're more obvious to people unfamiliar with the tools; please change them back unless there's a good reason not to :+1:

scripts/test.js
```diff
@@ -4,6 +4,27 @@ require("@app/config/env");
...
 const concurrently = require("concurrently");
 
+// Dear graphile-migrate, please treat the test DB as if it were the shadow DB
+process.env.SHADOW_DATABASE_URL = process.env.TEST_DATABASE_URL;
+
+// Signal to graphile-migrate scripts that we're in the tests
+process.env.IN_TESTS = "1";
+process.env.NODE_ENV = "test";
+
+const cmdArgs = process.argv;
```
MegaCookie, line 14: I changed this file and the way how arguments are handled to add the `--delay` argument. It should work fine, but I am not sure if this is the preferred way. Otherwise we could use [minimist](https://www.npmjs.com/package/minimist) package to handle command line arguments in a slightly more sophisticated way.

benjie, line 14: The first two args are `node` and the script name, so we should skip those

```suggestion
const cmdArgs = process.argv.slice(2);
```

scripts/test.js
```diff
@@ -4,6 +4,27 @@ require("@app/config/env");
...
+const watchMode = cmdArgs.find(
+  (arg) => arg === "--watch" || arg === "--watchAll"
+);
+const delayArg = cmdArgs.indexOf("--delay");
+let delaySeconds = null;
+if (delayArg > 0) {
+  delaySeconds = parseInt(cmdArgs[delayArg + 1]);
```
benjie, line 21: When using `parseInt` **always** pass the radix:

```suggestion
  delaySeconds = parseInt(cmdArgs[delayArg + 1], 10);
```

benjie: Actually, let's use parseFloat here so you could do e.g. `0.25` seconds

```suggestion
  delaySeconds = parseFloat(cmdArgs[delayArg + 1]);
```

scripts/test.js
```diff
@@ -47,48 +68,40 @@ const spawnSync = (cmd, args, options) => {
...
-} else {
-  // Run once, so just run the tests
-  spawnSync("jest", ["-i", ...process.argv.slice(2)], opts);
+        killOthers: ["failure"],
+      }
+    );
+  } else {
+    // Run once, so just run the tests
+    const argsWithoutDelayArg = process.argv
+      .slice(2)
```
benjie, line 103:
```suggestion
    const argsWithoutDelayArg = cmdArgs
```

benjie (changes requested):
Looks good; I'm happy with this approach. @ me when you're ready for a full re-review :raised_hands:

scripts/test.js
benjie: This is fine for now. No need for the extra weight of minimist for a couple flags.

MegaCookie:
> 
> 
> Looks good; I'm happy with this approach. @ me when you're ready for a full re-review 🙌

Hey Benjie, can you look at it if you have time? :)

@app/server/package.json
```diff
@@ -5,10 +5,10 @@
   "scripts": {
     "build": "tsc -b",
     "start": "node -r @app/config/env dist/index.js",
-    "dev": "nodemon --signal SIGINT --watch 'dist/**/*.js' -x 'node --inspect=9678 -r @app/config/env -r source-map-support/register' dist/index.js",
-    "schema:export": "node -r @app/config/env node_modules/.bin/ts-node -O '{\"rootDir\":null}' scripts/schema-export.ts",
+    "dev": "nodemon --signal SIGINT --watch 'dist/**/*.js' -x \"node --inspect=9678 -r @app/config/env -r source-map-support/register\" dist/index.js",
+    "schema:export": "npx --no-install -n \"-r @app/config/env\" ts-node -O '{\"rootDir\":null}' scripts/schema-export.ts",
```
benjie, line 9: I notice in the 'dev' line you've replaced `'` with `\"`; but the `-O` argument doesn't have the same change for `schema:export`. Is replacing `'` required? If so: why is it not required here, or was it simply missed? If not: we should go back to `'` because it's simpler.

package.json
```diff
@@ -12,12 +12,16 @@
     "pretest": "lerna run pretest",
     "test": "node scripts/test.js",
     "posttest": "lerna run posttest",
-    "test:watch": "node scripts/test.js --watch",
-    "lint": "tsc -b && yarn prettier:check && eslint --ext .js,.jsx,.ts,.tsx,.graphql .",
-    "lint:fix": "eslint --ext .js,.jsx,.ts,.tsx,.graphql . --fix; prettier --ignore-path .eslintignore --write '**/*.{js,jsx,ts,tsx,graphql,md}'",
-    "prettier:check": "prettier --ignore-path .eslintignore --check '**/*.{js,jsx,ts,tsx,graphql,md}'",
-    "depcheck": "lerna exec --stream 'yarn depcheck --ignores=\"@app/config,@app/client,tslib,webpack,babel-plugin-import,source-map-support,@graphql-codegen/*,graphql-toolkit,net,tls,dayjs\" --ignore-dirs=\".next\"'",
-    "dev": "yarn && lerna run codegen --stream && tsc -b && concurrently --kill-others --names 'TSC,WATCH,RUN,TEST' --prefix '({name})' --prefix-colors 'yellow.bold,yellow.bold,cyan.bold,greenBright.bold' 'tsc -b --watch --preserveWatchOutput --assumeChangesOnlyAffectDirectDependencies' 'lerna run --parallel watch' 'lerna run --parallel dev' 'sleep 10; yarn test:watch'",
+    "test:watch": "node scripts/test.js --watch --delay 10",
```
benjie, line 15: Move the delay back into the `dev` command below

```suggestion
    "test:watch": "node scripts/test.js --watch",
```

package.json
```diff
@@ -12,12 +12,16 @@
...
+    "lint": "tsc -b && yarn prettier:check && yarn eslint:check",
+    "lint:fix": "run-script-os",
+    "lint:fix:windows": "yarn eslint:check --fix & yarn prettier:fix",
+    "lint:fix:default": "yarn eslint:check --fix; yarn prettier:fix",
+    "eslint:check": "eslint --ext .js,.jsx,.ts,.tsx,.graphql .",
+    "prettier:check": "prettier --ignore-path .eslintignore --check \"**/*.{js,jsx,ts,tsx,graphql,md}\"",
+    "prettier:fix": "prettier --ignore-path .eslintignore --write \"**/*.{js,jsx,ts,tsx,graphql,md}\"",
+    "depcheck": "lerna exec --stream \"yarn depcheck --ignores=\"@app/config,@app/client,tslib,webpack,babel-plugin-import,source-map-support,@graphql-codegen/*,graphql-toolkit,net,tls,dayjs\" --ignore-dirs=\".next\"\"",
+    "dev": "yarn && lerna run codegen --stream && tsc -b && concurrently --kill-others --names \"TSC,WATCH,RUN,TEST\" --prefix \"({name})\" --prefix-colors \"yellow.bold,yellow.bold,cyan.bold,greenBright.bold\" \"tsc -b --watch --preserveWatchOutput\" \"lerna run --parallel watch\" \"lerna run --parallel dev\" \"yarn test:watch\"",
```
benjie, line 24:
```suggestion
    "dev": "yarn && lerna run codegen --stream && tsc -b && concurrently --kill-others --names \"TSC,WATCH,RUN,TEST\" --prefix \"({name})\" --prefix-colors \"yellow.bold,yellow.bold,cyan.bold,greenBright.bold\" \"tsc -b --watch --preserveWatchOutput\" \"lerna run --parallel watch\" \"lerna run --parallel dev\" \"yarn test:watch --delay 10\"",
```

package.json
```diff
@@ -12,12 +12,16 @@
...
-    "depcheck": "lerna exec --stream 'yarn depcheck --ignores=\"@app/config,@app/client,tslib,webpack,babel-plugin-import,source-map-support,@graphql-codegen/*,graphql-toolkit,net,tls,dayjs\" --ignore-dirs=\".next\"'",
-    "dev": "yarn && lerna run codegen --stream && tsc -b && concurrently --kill-others --names 'TSC,WATCH,RUN,TEST' --prefix '({name})' --prefix-colors 'yellow.bold,yellow.bold,cyan.bold,greenBright.bold' 'tsc -b --watch --preserveWatchOutput --assumeChangesOnlyAffectDirectDependencies' 'lerna run --parallel watch' 'lerna run --parallel dev' 'sleep 10; yarn test:watch'",
+    "test:watch": "node scripts/test.js --watch --delay 10",
+    "lint": "tsc -b && yarn prettier:check && yarn eslint:check",
+    "lint:fix": "run-script-os",
+    "lint:fix:windows": "yarn eslint:check --fix & yarn prettier:fix",
+    "lint:fix:default": "yarn eslint:check --fix; yarn prettier:fix",
+    "eslint:check": "eslint --ext .js,.jsx,.ts,.tsx,.graphql .",
+    "prettier:check": "prettier --ignore-path .eslintignore --check \"**/*.{js,jsx,ts,tsx,graphql,md}\"",
+    "prettier:fix": "prettier --ignore-path .eslintignore --write \"**/*.{js,jsx,ts,tsx,graphql,md}\"",
```
benjie, line 22: I think we can refactor this so that we extract the path spec parts. This also allows users to run `yarn eslint path/to/directory/` and have it do what they want (respecting all the extensions). `yarn prettier path/to/file.ts` will also work, which is why I've named the script that includes all the paths in the project `prettier:all`.

```suggestion
    "lint": "tsc -b && yarn prettier:all --check && yarn eslint .",
    "lint:fix": "run-script-os",
    "lint:fix:windows": "yarn eslint --fix . & yarn prettier:all --write",
    "lint:fix:default": "yarn eslint --fix .; yarn prettier:all --write",
    "eslint": "eslint --ext .js,.jsx,.ts,.tsx,.graphql",
    "prettier:all": "prettier --ignore-path .eslintignore \"**/*.{js,jsx,ts,tsx,graphql,md}\"",
```

benjie: Despite what I said previously I think I'm okay with changing to simply only run prettier if ESLint passes:

```
  "lint:fix": "yarn eslint --fix . && yarn prettier:all --write"
```

if this removes the need for `run-script-os`. It's a bit redundant to have three lines just so that we can use `&` / `;`. Sorry for the back-and-forth on this.

scripts/test.js
```diff
@@ -4,90 +4,104 @@ require("@app/config/env");
...
+    const result = rawSpawnSync(cmd, args, {
+      stdio: ["pipe", "inherit", "inherit"],
+      env: {
+        ...process.env,
+        YARN_SILENT: "1",
+        npm_config_loglevel: "silent",
+      },
+      shell: true,
+      ...options,
+    });
```
benjie, line 41: There's no need for this function to be defined in this scope (I think?), please move it back to the root scope.

yarn.lock
```diff
@@ -358,18 +391,30 @@
...
   dependencies:
     "@babel/types" "^7.11.0"
 
+"@babel/helper-split-export-declaration@^7.8.3":
+  version "7.8.3"
+  resolved "https://registry.yarnpkg.com/@babel/helper-split-export-declaration/-/helper-split-export-declaration-7.8.3.tgz#31a9f30070f91368a7182cf05f831781065fc7a9"
+  integrity sha512-3x3yOeyBhW851hroze7ElzdkeRXQYQbFIb7gLK1WQYsw2GWDay5gAJNw1sWJ0VFP6z5J1whqeXH/WCdCjZv6dA==
+  dependencies:
+    "@babel/types" "^7.8.3"
+
```
benjie, line 407: There's quite a few duplicates in `yarn.lock`; to resolve, please:

```
npx yarn-deduplicate && yarn
```

benjie (changes requested):
This is really coming together. The vast majority of these changes are now very straightforward and easy to justify. I've left a few comments, and it would be helpful if you could try and minimize the diff in `scripts/test.js` so that it only shows the necessary changes when "Hide whitespace changes" is turned on (`?w=1`), but I think this is very close to being mergeable.

Thanks again for your stellar work on this :raised_hands:

package.json
benjie: (See next comment)

scripts/test.js
MegaCookie: It was necessary as it was defined as it could not find the anonymous function in the main scope. But it also works if we change to a normal function declaration.

@app/server/package.json
MegaCookie: Indeed it was not working on Windows as there everything needs to be double quotes in the package.json. So with all the escaping we need to use 6 backslashes unfortunately. It was the only way I could make it work

benjie (changes requested):
Just this one backslash issue remaining (you knew it was coming!); if we have to have the 7 backslashes then so be it, but I'd like to understand a) how it works on Windows, and b) how it works on Unix. I've suggested an alternative, if that works I think we're good to merge.

@app/server/package.json
```diff
@@ -6,7 +6,7 @@
     "build": "tsc -b",
     "start": "node -r @app/config/env dist/index.js",
     "dev": "nodemon --signal SIGINT --watch 'dist/**/*.js' -x \"node --inspect=9678 -r @app/config/env -r source-map-support/register\" dist/index.js",
-    "schema:export": "npx --no-install -n \"-r @app/config/env\" ts-node -O '{\"rootDir\":null}' scripts/schema-export.ts",
+    "schema:export": "npx --no-install -n \"-r @app/config/env\" ts-node -O {\\\\\\\"rootDir\\\\\\\":null} scripts/schema-export.ts",
```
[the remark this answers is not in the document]
MegaCookie: Yup that option is way better to use :)

benjie:
:sparkles: If that works; it's delicious.

It's going to be a fair while before I can review this fully again (and test nothing has broken) but I think it should be an easy merge once I've validated it. Thanks for all your hard work on this! :raised_hands:

benjie (approved):
I need to test this; but it looks good to me!

benjie:
Still works perfectly on Linux so I'm going to merge it; if it doesn't work on Windows then people are welcome to follow up with specific fixes :+1:

benjie:
Thanks @MegaCookie this is finally merged! :raised_hands: :tada:

Issue #205, opened 2020-10-02
Incorporate postgraphile-tag-json-tools


benjie:
### Feature description

@phryneas has written https://github.com/mayflower/postgraphile-tag-json-tools which enables more accurate JSON schema for the postgraphile.tags.jsonc file, leading to better auto-complete (and hopefully detection of misconfiguration, e.g. typos in references, invalid tags, etc).

### Breaking changes

Potentially more esoteric tags won't be supported.

PR #206, opened 2020-10-07
feature: add postgraphile-tag-json-tools for json schema generation

phryneas:
Like requested in #205 

## Description

This PR only uses the `createJsonSchema` part of `postgraphile-tag-json-tools`, as `extractSmartTags` and `loadSmartTags` functionalities do not seem necessary here.

Changes in the database will update `@app/server/pg-database-smart-tags.schema.json`, but edits in `@app/server/pg-smart-tags.schema.json` will persist, so additional tags can be added without fear of something being overridden. (I added simpleCollections and resultFieldName, as I didn't have those in my initial schema - if you can think of any more tags, please mention them & I will add them.)

I've marked the place to add new tags with `"$comment": "------ Add extra tags here, in `properties`. ------",`.

## Performance impact

None in production use, probably neglectable in dev mode.

## Security impact

None in production use, nothing I could think of in dev use.

## Checklist

### DIFF
--- a/@app/server/src/middleware/installPostGraphile.ts
+++ b/@app/server/src/middleware/installPostGraphile.ts
@@ -12,6 +12,7 @@ import {
   postgraphile,
   PostGraphileOptions,
 } from "postgraphile";
+import { createJsonSchema as CreateJsonSchemaPlugin } from "postgraphile-tag-json-tools";
 import { makePgSmartTagsFromFilePlugin } from "postgraphile/plugins";
 
 import { getHttpServer, getWebsocketMiddlewares } from "../app";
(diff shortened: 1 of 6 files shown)


@app/server/src/middleware/installPostGraphile.ts
```diff
@@ -183,6 +184,14 @@ export function getPostGraphileOptions({
...
       // Adds custom orders to our GraphQL schema
       OrdersPlugin,
+
+      // Plugins for just Development mode
+      ...(!isDev
+        ? []
+        : [
+            // updates the `pg-database-smart-tags.schema.json` file based on the database
+            CreateJsonSchemaPlugin,
+          ]),
```
benjie, line 194: Avoiding the negative makes this easier to read IMO.

```suggestion
      ...(isDev
        ? [
            // updates the `pg-database-smart-tags.schema.json` file based on the database
            CreateJsonSchemaPlugin,
          ]
        : []),
```

benjie (changes requested):
This is really cool! It could do with a smidge of docs; could you put some in the `@app/server/README.md` file referencing the two newly created file names and a link to postgraphile-tag-json-tools please?

PR #208, opened 2020-10-22
Update installSession.ts

hegelstad:
## Description

Warn user that sessions are not being used due to requests not originating from same origin.

Suggestions to make this more actionable and self-explanatory is appreciated. I think the main problem this solves is that new users doesn't have to search forever to figure out why their auth isn't working. Of course everyone should read the docs, but this can aid new users in adopting postgraphile.

### DIFF
--- a/@app/server/src/middleware/installSession.ts
+++ b/@app/server/src/middleware/installSession.ts
@@ -78,6 +78,7 @@ export default (app: Express) => {
     if (req.isSameOrigin) {
       sessionMiddleware(req, res, next);
     } else {
+      console.warn('⚠️ Request did not come from SameOrigin, session was not created.')
       next();
     }
   };


benjie (changes requested):
Thanks for raising this PR; it's important to help people find the mistakes in their code quickly; especially within a big complex system such as Starter.

There's perfectly valid reasons to not have the request come from SameOrigin (e.g. if you have a mobile app, or are loading resources from a different subdomain), so outputting this for every request would be a bit much. Maybe you could add a few enhancements:

1. Only do this in development mode
2. Track the incorrect origin and only output the message once per origin
3. Include more details in the warning: what were the expected and actual origins?
4. Add a comment that makes it clear that it's fine to delete this message if you don't need it; maybe wrap it with a constant like `if (WARN_ON_DIFFERENT_ORIGIN) { ... }`

PR #210, opened 2020-10-29
Update heroku-setup.template

hegelstad:
placeholder name should reflect the Heroku naming conventions.

## Checklist


benjie:
Thanks for your PR, but this is deliberately all-caps to solicit attention and make it clear it needs to be replaced. The comment above it states "Must start with a letter, end with a letter or digit and can only contain **lowercase letters**, digits, and dashes." which I believe is sufficient to convey this concern.

PR #211, opened 2020-10-29
Add --region flag to heroku apps:create in heroku-setup.template

hegelstad:
Add region parameter to dyno creation. Changing the region is non-trivial so better include it to make it easier to get a nice default config.

## Checklist


benjie (approved):
Good idea :+1:

PR #212, opened 2020-11-10
chore: upgrade antd and fix `useForm` import

jackbravo:
I'm getting this error on `http://localhost:5678/create-organization/`

```
Unhandled Runtime Error TypeError: Object is not a function or its return value is not iterable
```

## Checklist


jackbravo:
@benjie have you not run into this issue?

benjie:
Works fine for me on the version of AntD the starter has in `yarn.lock`:

Peek 2020-11-20 16-28


Further we have integration tests that check this flow works: https://github.com/graphile/starter/blob/main/%40app/e2e/cypress/integration/organization_create.spec.ts

This PR doesn't add the updated `yarn.lock`; but I'm guessing if I updated antd it'll start failing.

benjie:
Can confirm it starts to fail when you upgrade AntD.

benjie (approved):
Thanks :raised_hands:

benjie:
Registration fails, as can be seen in the Cypress tests. Reproducing locally and terminal says:

Screenshot_20201120_174522

benjie:
I got as far as figuring out that it's failing because `outOfDate` is true; but no further

Screenshot_20201120_185127


Screenshot_20201120_185106

jackbravo:
Hi Benji!

Ok, I'll try to take a look at this today. Thanks!

benjie:
Turns out the way we were keeping the password validations in sync was not compatible with the upgraded version. I've fixed it.

PR #214, opened 2020-11-18
chore(ci): upgrade cypress-io/github-action@v1 to v2

jackbravo:
## Description

Upgrade cypress-io/github-action@v1 to v2 to avoid `set-env` command is disabled error that appears on github status checks.

GitHub is deprecating set-env and add-path commands, causing the cypress end-to-end status check to log the following warning when executed:

> The set-env command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

## Checklist

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Start worker in background
         run: yarn worker start &
       - name: Cypress run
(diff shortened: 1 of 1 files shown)


benjie (approved):
Awesome; thanks :raised_hands:

PR #219, opened 2020-11-27
feat(handleErrors): better error for foreign key violations

lukeramsden:
After this change we can determine errors for foreign key violations on the frontend.

```ts
const code = getCodeFromError(e);
const exception = getExceptionFromError(e);
const fields: any = exception && exception["fields"];
// This PR creates BADFK error code, and fixes `fields` to work with fkeys
if (code === "BADFK" && fields && fields[0] === "genre") {
  helpers.setFieldError(
    "genre",
    "That genre does not exist in our database. Please pick an existing one, or don't provide a genre at all."
  );
} else {
  helpers.setStatus({
    error: true,
    message: e?.message || "Unknown error",
  });
}
```

I don't see anywhere in the current codebase this would be used or that this would affect, although I may have missed something. It's more a case of expected behaviour than specifically a fix for this project front-end.

### DIFF
--- a/@app/server/src/utils/handleErrors.ts
+++ b/@app/server/src/utils/handleErrors.ts
@@ -60,7 +60,7 @@ export const ERROR_MESSAGE_OVERRIDES: { [code: string]: typeof pluck } = {
   }),
   "23503": (err) => ({
     ...pluck(err),
-    message: "Invalid foreign key",
+    message: "Invalid reference",
     fields: conflictFieldsFromError(err),
     code: "BADFK",
   }),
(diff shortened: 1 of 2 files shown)


@app/server/src/utils/handleErrors.ts
```diff
@@ -58,20 +58,31 @@ export const ERROR_MESSAGE_OVERRIDES: { [code: string]: typeof pluck } = {
     fields: conflictFieldsFromError(err),
     code: "NUNIQ",
   }),
+  "23503": (err) => ({
+    ...pluck(err),
+    message: "Invalid foreign key",
```
benjie, line 63:
```suggestion
    message: "Invalid reference",
```

benjie (approved):
Thanks :+1: 

Note this is still hacky (like the current implementation for conflicts was) - really we should get the fields from the database introspection of the constraint name, and we should use inflection to rename them rather than camelCase. However... that's a lot of work... so :grimacing: :+1:

PR #221, opened 2020-11-29
Update express-session and add custom returnTo property

jackbravo:
## Description

Also had some errores while updating express-session to latest release.

Specifically because of this change:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express-session/index.d.ts#L202

from:

Reported here: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/49484

## Checklist

### DIFF
--- a/@app/client/src/next.config.js
+++ b/@app/client/src/next.config.js
@@ -37,7 +37,7 @@ if (!process.env.ROOT_URL) {
     )({
       poweredByHeader: false,
       distDir: `../.next`,
-      exportTrailingSlash: true,
+      trailingSlash: false,
       lessLoaderOptions: {
         javascriptEnabled: true,
         modifyVars: themeVariables, // make your antd custom effective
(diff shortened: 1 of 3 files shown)


benjie:
It seems the updates to the lockfile have caused cascading failures in the cypress tests. Apparently someone's not been adhering to semver :/ Those issues will need fixing before this can be merged; but I think that's probably beyond the scope of this PR. I'm going to leave this open until I have time to address this breakage, but thank you for sharing the fix :+1:

lukeramsden:
https://github.com/vercel/next.js/pull/15768

Could be this?

lukeramsden:
`exportTrailingSlash: true` -> `trailingSlash: false` seems to fix the issue on my starter-derived project, maybe one of you could make that change on this PR and see if that fixes the E2E tests?

jackbravo:
It worked @lukeramsden ! 🎉 

@benjie this seems ready now.

benjie:
Awesome; merged into #222

benjie:
Sorry for the delay; took a long time to track down a bug that turned out to be `next build` not working quite right in `NODE_ENV=test`

PR #222, opened 2020-11-30
chore(deps): upgrade most dependencies

benjie:
## Description

Everything except Apollo Client, which needs a bit more work.

## Performance impact

Unknown.

## Security impact

Unknown.

---

Fixes #224 
Fixes #223
Fixes #221

PR #224, opened 2020-12-07
Upgrades joaquin

jackbravo:
## Description

Attempts to fix cypress [browser crash.](https://github.com/cypress-io/cypress/issues/350)

Using fix proposed at: https://github.com/cypress-io/github-action/issues/104#issuecomment-666047965

## Checklist


jackbravo:
seems to not be running the cypress CI check, so this won't work 😛 .

benjie:
Thanks for keeping chipping away at this @jackbravo. I've no idea what's up or even if it's Cypress' fault or if the issue lies elsewhere and Cypress is paying for it.

jackbravo:
@benjie I feel like I need to install a virtual machine with ubuntu 16.04 to test cypress locally. This flow of testing is very slow :-p. And sadly my work machine today is a mac.

benjie:
I'm running Ubuntu LTS 20.04 and can't reproduce here. I wonder if upgrading `main` branch's Cypress only would have the same issues?

jackbravo:
I know that ubuntu-latest on github-actions means ubuntu 16.04. Maybe we
could try using a newer ubuntu, but cypress is supposed to work on all
versions:
https://github.com/cypress-io/github-action/issues/251

mattvb91:
@jackbravo did you manage to make any progress with the out of memory? I upgraded our actions to Ubuntu 20.04 and still experiencing the same issue

jackbravo:
@mattvb91 sadly I don't have made any progress, and haven't got the time to investigate other options.

benjie:
I also attempted to fix it in https://github.com/graphile/starter/pull/222 but no luck. Can anyone reproduce this outside of GitHub actions?

benjie:
I've reproduced it locally! To do so was quite simple in the end:

```
NODE_ENV=test yarn build
NODE_ENV=test yarn server start &
NODE_ENV=test yarn worker start &
yarn e2e run run --browser chrome
```

Turns out we can't even load the browser in this circumstance either; the issue seems to be Next.js related rather than Cypress related! :tada:

benjie:
Fixed in [`3afcb19` (#222)](https://github.com/graphile/starter/pull/222/commits/3afcb19d85c399ac1cddca8187ba63f9ead02796)

benjie:
Thanks for your help debugging this everyone! Turned out to be `next build` not working quite right in `NODE_ENV=test`

PR #230, opened 2021-01-09
fix: remove npx -n usage in favor of cross-env / NODE_OPTIONS

DylanVann:
## Description

On cloning this repo and following instructions I encountered errors saying `-n` is no longer supported in `npx`.

image

It seems that my version of `npx` is too old to work correctly. I normally use [`fnm`](https://github.com/Schniz/fnm) to manage Node versions and don't use `npm` or `npx` too often so I hadn't noticed this.

More detail: I've installed Node `15.5.1` using `fnm`, this installs `npm@7.3.0` and `npx@7.3.0`, I searched online but I don't see any matrix mapping `npm version -> npx version` anywhere, maybe `fnm` is doing something incorrect by using the same version as `npm`.

This could be mitigated for other users by using `NODE_OPTIONS` instead of `npx`.

## Performance impact

Possibly slightly better perf 🤷‍♂️ , most of these commands were already using `cross-env`.

## Security impact

None.

## Checklist

- [x] I have added this feature to 'Pending' in the `RELEASE_NOTES.md` file (if one exists).

### DIFF
--- a/@app/db/package.json
+++ b/@app/db/package.json
@@ -14,11 +14,11 @@
     "test": "cross-env NODE_ENV=test NODE_OPTIONS=\"$NODE_OPTIONS -r @app/config/env\" jest"
   },
   "dependencies": {
+    "cross-env": "^7.0.2",
     "graphile-migrate": "^1.0.1"
   },
   "devDependencies": {
     "@types/pg": "^7.14.4",
-    "cross-env": "^7.0.2",
     "graphile-worker": "^0.8.1",
     "jest": "^26.6.3",
     "lodash": "^4.17.15",
--- a/@app/db/package.json
+++ b/@app/db/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "version": "0.0.0",
   "scripts": {
-    "gm": "npx --no-install -n \"-r @app/config/env\" graphile-migrate",
+    "gm": "cross-env NODE_OPTIONS=\"$NODE_OPTIONS -r @app/config/env\" graphile-migrate",
     "migrate": "yarn gm migrate",
     "watch": "yarn gm watch",
     "commit": "yarn gm commit",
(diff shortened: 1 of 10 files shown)


benjie:
Do you know if this still works on Windows? That’s why we had to do the npx stuff iirc.

DylanVann:
I believe it should because it's making use of `cross-env`. `NODE_OPTIONS` should work on Windows.

After trying to get a Windows dev environment working in VirtualBox for over 2 hours I'm calling it quits, maybe someone else can test.

Not sure what it does exactly, but the fact that "Windows Node CI" is passing seems to be a good sign.

benjie:
Awesome; thanks for checking @DylanVann. We seem to be having some CI issue but they look transient, so I'll trigger a rebuild and see if we can get them to pass.

@app/worker/package.json
```diff
@@ -3,12 +3,12 @@
   "private": true,
   "version": "0.0.0",
   "scripts": {
-    "gw": "cd dist && npx --no-install -n \"-r @app/config/env\" graphile-worker",
+    "gw": "cd dist && cross-env NODE_OPTIONS=\"$NODE_OPTIONS -r @app/config/env\" graphile-worker",
```
benjie, line 6: To use `cross-env` here, we need to move it from `devDependencies` to `dependencies`.

@app/db/package.json
```diff
@@ -3,15 +3,15 @@
   "private": true,
   "version": "0.0.0",
   "scripts": {
-    "gm": "npx --no-install -n \"-r @app/config/env\" graphile-migrate",
+    "gm": "cross-env NODE_OPTIONS=\"$NODE_OPTIONS -r @app/config/env\" graphile-migrate",
```
benjie, line 6: To use cross-env here we should move it to `dependencies`.

benjie (changes requested):
Looks good; the main issue is the production docker build fails because the worker fails to start (so cannot be torn down); I believe this is because `cross-env` needs to be a production dependency within the worker now. We ought to do the same for `@app/db`.

DylanVann:
That did the trick, thanks!

benjie:
Awesome; thanks! :raised_hands:

PR #234, opened 2021-01-14
chore: refactor environment variables

DylanVann:
## Description

This is refactors environment variables so that all validation of them is done in one location (`@app/config/src/index.ts`).

This is only used by the server package in this PR, if accepted a followup can change the usage in other packages as well.

## Performance impact

In the [Express docs](https://expressjs.com/en/advanced/best-practice-performance.html):

> If you need to write environment-specific code, you can check the value of NODE_ENV with process.env.NODE_ENV. Be aware that checking the value of any environment variable incurs a performance penalty, and so should be done sparingly.

Some [notes about this in relation to React SSR](https://github.com/nodejs/node/issues/3104#issuecomment-161430881).

I haven't profiled as I don't think it would make a big difference, but that makes some intuitive sense.

## Security impact

Developers need to be careful not to import secrets from `@app/config` in client code, or their secrets could get leaked, I believe Next.js might prevent this though, by not passing through environment variables to the browser. I can check this

## Checklist

- [x] I have added this feature to 'Pending' in the `RELEASE_NOTES.md` file (if one exists).


@app/config/src/index.ts
```diff
@@ -3,11 +3,170 @@ const packageJson = require("../../../package.json");
...
+  12
+).value;
+export const GRAPHQL_COST_LIMIT: number = defaultNumber(
+  parseInteger({
+    value: process.env.GRAPHQL_COST_LIMIT,
+    name: "GRAPHQL_COST_LIMIT",
+  }),
+  30000
+).value;
+export const HIDE_QUERY_COST: boolean =
```
DylanVann, line 136: I don't really understand what this one was doing in the original code, but I think this functions the same.

@app/config/src/index.ts
```diff
@@ -3,11 +3,170 @@ const packageJson = require("../../../package.json");
...
-export const emailLegalText =
+export const fromEmail: string = `"PostGraphile Starter" <no-reply@examples.graphile.org>`;
+export const awsRegion: string = "us-east-1";
+export const projectName: string = packageJson.name.replace(/[-_]/g, " ");
+export const companyName: string = projectName; // For copyright ownership
+export const emailLegalText: string =
   // Envvar here so we can override on the demo website
   process.env.LEGAL_TEXT || "<Insert legal email footer text here >";
+
+const required = (v: {
```
DylanVann, line 14: A mini validation framework:
- `required`
- `defaultNumber`
- `minLength`
- `parseInteger`

Other options were using `envalid`, `io-ts`, etc. I think it's simple enough that isn't needed.

`name` is passed instead of being used to access the environment variables because some bundlers will replace `process.env.X` with the value, which would not work if we access `process.env` dynamically.

@app/server/src/middleware/installSSR.ts
```diff
@@ -1,17 +1,9 @@
...
 
-if (!process.env.NODE_ENV) {
-  throw new Error("No NODE_ENV envvar! Try `export NODE_ENV=development`");
-}
-
-const isDev = process.env.NODE_ENV === "development";
-
 export default async function installSSR(app: Express) {
-  // @ts-ignore Next had a bad typing file, they claim `export default` but should have `export =`
-  // Ref: https://unpkg.com/next@9.0.3/dist/server/next.js
```
DylanVann, line 14: This and the comment at the top of this file seem to have been resolved. Also `// Foo` below may have been resolved 🤔 .

scripts/_setup_utils.js
```diff
@@ -72,17 +72,24 @@ exports.updateDotenv = function updateDotenv(add, answers) {
...
   add(
-    "JWT_SECRET",
+    "SECRET_JWT_ACCESS",
+    safeRandomString(48),
+    `\
+# This secret is used for signing JWT access tokens (we don't use this by default)`
+  );
+
+  add(
+    "SECRET_JWT_REFRESH",
```
DylanVann, line 89: If an app is using a JWT strategy usually it needs to encode `accessToken`s and `refreshTokens`.

DylanVann:
I don't know that much about the docker setup. A hint about how to make sure the docker tests get `SECRET` and `JWT_SECRET` (which are now validated/required) would be appreciated 😄 

Doesn't the docker version currently need the `SECRET` to function correctly? The results seem to be saying that it is not defined.

@app/config/src/validate.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Validation functions for environment variables.
+ *
+ * `name` is passed with `value` so that bundlers that replace
+ * process.env.MY_VAR with the value will still provide useful error messages.
```
DylanVann, line 5: A mini validation framework:
- `required`
- `defaultNumber`
- `minLength`
- `parseInteger`

Other options were using `envalid`, `io-ts`, etc. I think it's simple enough that isn't needed.

@app/config/src/index.ts
```diff
@@ -1,13 +1,121 @@
...
-  '"PostGraphile Starter" <no-reply@examples.graphile.org>';
-export const awsRegion = "us-east-1";
-export const projectName = packageJson.name.replace(/[-_]/g, " ");
-export const companyName = projectName; // For copyright ownership
-export const emailLegalText =
+export const fromEmail: string = `"PostGraphile Starter" <no-reply@examples.graphile.org>`;
+export const awsRegion: string = "us-east-1";
+export const projectName: string = packageJson.name.replace(/[-_]/g, " ");
+export const companyName: string = projectName; // For copyright ownership
+export const emailLegalText: string =
```
benjie, line 12: I'm surprised the linter does not complain about unnecessary type definitions here (since they can be implied by the values).

benjie (changes requested):
Interesting approach. I'm concerned about this being easily misused; could you move most of the new variables to a new path `@app/config/server` so that it makes it very obvious it should not be required in the client? (You'll may need to create a `server.js` file that just contains `module.exports = require('./dist/server.js')` so you can write a `src/server.ts`. I'm not ready to use the fancy `package.json` module mapping features yet.)

This isn't a full review.

benjie:
This approach might not be the best since it enforces variables whenever `process.env` would have been referenced, this is what is causing the Docker issues. The docker image should be able to build just fine without the `SECRET` being provided - the `SECRET` should only be required at runtime (whereas other variables like `NODE_ENV` are required at build time). Switching it to an "on demand" model where the first time an envvar is referenced it's validated and cached might be a better approach.

Another concern is that it references all envvars in any place any of them are referenced, whereas the previous code only references the required envvars in each place (so if you just run the worker you only read the envvars relevant to the worker). This may or may not have an effect on webpacking your server/etc. It will have an effect if you only expose the envvars each service needs to that service (e.g. the server has different envvar requirements to the worker, in an advanced deployment you might like to separate these out but this would prevent you doing that).

DylanVann:
Thank you for the feedback! I'll take a second attempt at this soon.

I think you're right that this ties too many things together. The different components should only validate the environment variables they need.

@app/config/src/index.ts
DylanVann: I'd usually go for inferred but I wanted to ensure that people changing the validators didn't accidentally do something like change the exported type from `string` to `string | undefined` without making it explicit.

benjie:
_[semi-automated message]_ Hi, there has been no activity in this issue for a while so I'm closing it to keep the issues/pull requests manageable. If this is still an issue, please re-open with additional details.

Issue #237, opened 2021-01-21
Empty white screen in Safari, due to Content Security Policy


singingwolfboy:
### Summary

Safari apparently has a more restrictive default [content security policy](https://content-security-policy.com) than Chrome or Firefox. This prevents the starter from displaying properly (or at all) in Safari.

### Steps to reproduce

Visit https://graphile-starter.herokuapp.com using the latest version of Safari.

### Expected results

The website should appear, and be usable.

### Actual results

The website appears for a moment, but then is replaced with a blank white page. Opening the browser console reveals several security-related errors, that point to CSP violations.

<img width="1076" alt="safari-starter-broken" src="https://user-images.githubusercontent.com/132355/105372878-5cb72e00-5c06-11eb-8dbf-503d91391a56.png">

The most relevant errors are:

> [Error] Refused to execute a script because its hash, its nonce, or 'unsafe-inline' does not appear in the script-src directive of the Content Security Policy. (graphile-starter.herokuapp.com, line 0)
> [Error] Refused to connect to wss://graphile-starter.herokuapp.com/graphql because it appears in neither the connect-src directive nor the default-src directive of the Content Security Policy.

### Additional context

I am using Safari 14 on macOS Big Sur.

### Possible Solution

Set a `Content-Security-Policy` header in the response. [You can do this in the `next.config.js` file.](https://nextjs.org/docs/api-reference/next.config.js/headers)

* [MDN docs for Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP)
* [CSP quick reference guide](https://content-security-policy.com)

kh0r:
The wss error can be explained by the definition of 'self' on the corresponding [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src) for connect-src:
> Refers to the origin from which the protected document is being served, including the **same URL scheme** and port number.

Safari correctly refuses the connection via wss, because it is a different origin than the https url.

A possible fix would be to add the corresponding URL schemes to the [csp config here](https://github.com/graphile/starter/blob/cd9bc1ecc0a3867f3dd3a87803bf431f4c125723/%40app/server/src/middleware/installHelmet.ts):
(I also had to remove the `upgrade-insecure-requests` and  `block-all-mixed-content` directives to make the site work on http://localhost in Safari)
```
export default function installHelmet(app: Express) {
  const {
    ["upgrade-insecure-requests"]: _uir,
[...]

PR #244, opened 2021-01-29
fix(compat): fix CSP connect-src for Safari; force SSL option

benjie:
## Description

See #237; Safari doesn't seem to support `connect-src 'self'` referring to websocket protocol for websockets... So we have to list the full URL.

## Performance impact

Negligible.

## Security impact

There shouldn't be any, but it does relate to CSP.


SnowballAntrobus:
When using an unmodified starter , Safari still gives: Failed to load resource: An SSL error has occurred and a secure connection to the server cannot be made. When using Firefox or Chrome the site loads as expected.

benjie:
Are you using SSL locally?

Issue #247, opened 2021-02-07
Dependency requires Node 14 but Dockerfile defaults to Node 12


pellicceama:
I set up the project as per the instructions and tried to start the components inside of docker. I got the following error with the `postgraphile` library. 

```
server_1  | (RUN) @app/server: /work/node_modules/postgraphile/build-turbo/postgraphile/http/subscriptions.js:46
server_1  | (RUN) @app/server:     if (!websockets?.length) {
server_1  | (RUN) @app/server:                     ^
server_1  | (RUN) @app/server: SyntaxError: Unexpected token '.'
server_1  | (RUN) @app/server:     at wrapSafe (internal/modules/cjs/loader.js:915:16)
server_1  | (RUN) @app/server:     at Module._compile (internal/modules/cjs/loader.js:963:27)
server_1  | (RUN) @app/server:     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
server_1  | (RUN) @app/server:     at Module.load (internal/modules/cjs/loader.js:863:32)
server_1  | (RUN) @app/server:     at Function.Module._load (internal/modules/cjs/loader.js:708:14)
server_1  | (RUN) @app/server:     at Module.require (internal/modules/cjs/loader.js:887:19)
server_1  | (RUN) @app/server:     at require (internal/modules/cjs/helpers.js:74:18)
server_1  | (RUN) @app/server:     at Object.<anonymous> (/work/node_modules/postgraphile/src/postgraphile/http/createPostGraphileHttpRequestHandler.ts:54:1)
server_1  | (RUN) @app/server:     at Module._compile (internal/modules/cjs/loader.js:999:30)
server_1  | (RUN) @app/server:     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
server_1  | (RUN) @app/server: [nodemon] app crashed - waiting for file changes before starting...
```

Only Node 14+ supports [optional chaining](https://stackoverflow.com/a/59574160) which is required by `postgraphile`.

Changing the `Dockerfile` baseline image from `FROM node:12` to `FROM node:14` seems to do the trick. All the tests seem to pass and the apps boots up. I haven't performed any further testing. Will PR a fix, albeit this might require a fix in the library as well to be more defensive with node versions (not sure if possible).

benjie:
PostGraphile needs updating, this was a bug in one of the release candidates that I believe is now solved.

PR #248, opened 2021-02-07
chore(docker): upgrade to Node 14 base image

pellicceama:
Only Node 14+ supports optional chaining which is required by `postgraphile`.

This fixes https://github.com/graphile/starter/issues/247

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12
+FROM node:14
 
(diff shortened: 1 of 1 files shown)


benjie:
PostGraphile supports Node 8 and above; there was a bug in one of the release candidates but that should now be fixed.

None the less I don't see any reason not to upgrade to Node 14 :+1:

benjie (approved):

PR #254, opened 2021-03-09
chore: fix typos & remove duplicate text from 000001.sql

petrus-jvrensburg:
No changes to code... only updates to comments in committed migration file.


benjie (approved):
Thanks! :+1:

PR #271, opened 2021-09-04
feature: use zx to improve scripting dev experience

zacaytion:
## Description

Adds the [zx](https://github.com/google/zx) package to improve the experience of shell scripting with node. `zx` provides useful wrappers around `child_process`, escapes arguments and gives sensible defaults. 

ZX uses ES modules and prefers to use `.mjs` extension. If this is undesirable, the PR can be updated to use `.js` and wrap the scripts in `void async function () {}()`. The scripts could also be written in typescript as ZX [can compile and execute `.ts` files](https://github.com/google/zx#typescript-scripts)

## Performance impact

None, except when scripts are written in typescript. `.ts` files are passed to `ts-node` which is slower than calling `node` directly.

## Security impact

N/A || unknown

## Checklist


benjie:
Thanks for raising a pull request. Alas, I'm concerned that moving to shell commands, e.g. using `rm -rf` instead of rimraf, will decrease our Windows compatibility. I also see negligible benefit in adding this extra dependency and all its dependencies when it seems most of its functionality is already provided by node natively just with slightly different syntax. I'm going to leave the PR open because I think other people may find it useful (particularly people who only need to support unix... though if this were me I'd personally just use bash scripts directly) but I think it's unlikely I'll merge it.

PR #272, opened 2021-10-10
upgrade typescript

singingwolfboy:
## Description

Upgrade Typescript to 4.4.4, tslib to 2.3.1.

## Performance impact

None, as far as I know.

## Security impact

None, as far as I know.


heracek:
Hello, I have upgrade of TypeScript in PR #295

Problem with `catch(e)` was solved by simple `catch(e: any)`.

I know that it's not optimal, basically I'm giving up any type information about errors, byt it fits the requirement of being a minimal change in codebase.

Maybe type safe solution can be solved in different PR.

benjie:
We're now running TypeScript v5 as of #341

PR #275, opened 2021-10-16
Remove async/await that has no effect

caillou:
## Description

Remove async/await that has no effect.

## Performance impact

This should not change how the code runs.

## Security impact

None.

## Checklist

### DIFF
--- a/@app/client/src/pages/create-organization/index.tsx
+++ b/@app/client/src/pages/create-organization/index.tsx
@@ -34,10 +34,10 @@ const CreateOrganizationPage: NextPage = () => {
   const [slugCheckIsValid, setSlugCheckIsValid] = useState(false);
   const checkSlug = useMemo(
     () =>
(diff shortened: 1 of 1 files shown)


caillou:
So I took some more time to make sure my assumptions are right.

`lookupOrganizationBySlug` indeed returns `void` and there is no point `await`ing it, nor does it make sense to wrap it within a `try/catch` block.

It is a `LazyQuery`. The potential error can be found in the `slugError` variable.

As of now `setSlugCheckIsValid` is always executed before `lookupOrganizationBySlug` returns. No matter if you `await` it or not. Calling `setSlugCheckIsValid` sets `slugLoading` to `true`, which is why the loading indicator remains.

I have a new commit that completely removes the `try/catch` block. Let me know if I may open a new PR with this: https://github.com/caillou/starter/commit/de26a6f236c0dcdcd8695d07606dbf36807af392

Issue #278, opened 2021-10-28
can't launch database with docker-compose v2


DanielFGray:
### Summary

docker-compose 2 seems to behave slightly differently with regards to quotes in environment files

### Steps to reproduce

* install docker-compose v2
* clone starter repo
* `docker-compose up -d db`

### Expected results

```
 ⠿ Container graphile-starter-db-1           Started
```

### Actual results

unexpected character "-" in variable name near "exec -T db pg_dump\n\n# Drops tables without asking in \`yarn setup\`. Reasoning: 1) docker-compose is\n# not tty, 2) it's a dev env anyway.\nCONFIRM_DROP=y\n\n# POSTGRES_PASSWORD is the superuser password for PostgreSQL, it's required to\n# initialize the Postgres docker volume.\nPOSTGRES_PASSWORD=password\n\n# We're accessing Postgres via Docker, so we must use the db host and the\n# relevant password.\nDATABASE_HOST=db\nROOT_DATABASE_URL=postgres://postgres:password@db/postgres\n"

DanielFGray:
fixed with #279

PR #279, opened 2021-10-28
fix: quoting in docker yarn setup

DanielFGray:
## Description

fixes quoting in docker .env file for docker-compose v2, issue #278 

## Performance impact

unknown

## Security impact

unknown

## Checklist


DanielFGray:
> Will this still work with compose v1?
I havent figured out how to downgrade docker-compose on my system and test it myself, but I feel fairly confident it would

> Are double quotes an option?
My understanding of yaml is that they are interchangeable, a quick test on my machine doesn't yield any noticeable difference.

benjie (approved):

PR #283, opened 2021-11-28
fix: `/settings` page - incorrect errorcode checked

JonParton:
## Description

On the `/settings` page the error for the username being unique is not currently being handled correctly. 

The check on errors coming back from Apollo is checking for the error code `23505` but it is actually being rewritten as `NUNIQ` and so does not match. 

The error codes are stated in the [error_codes.md](/graphile/starter/blob/main/docs/error_codes.md) file, replaced using the [handleErrors.ts](/graphile/starter/blob/main/@app/server/src/utils/handleErrors.ts) file and I can also confirm I am physically seeing `NUNIQ` being returned when using the starter.

## Performance impact

None

## Security impact

None

## Checklist

### DIFF
--- a/@app/client/src/pages/settings/index.tsx
+++ b/@app/client/src/pages/settings/index.tsx
@@ -84,7 +84,7 @@ function ProfileSettingsForm({
         setSuccess(true);
       } catch (e) {
         const errcode = getCodeFromError(e);
-        if (errcode === "23505") {
+        if (errcode === "NUNIQ") {
           form.setFields([
             {
               name: "username",


@app/client/src/pages/settings/index.tsx
```diff
@@ -84,7 +84,7 @@ function ProfileSettingsForm({
         setSuccess(true);
       } catch (e) {
         const errcode = getCodeFromError(e);
-        if (errcode === "23505") {
+        if (errcode === "NUNIQ") {
```
benjie, line 87: Both errors can occur depending on at what stage the issue is detected, IIRC:

```suggestion
        if (errcode === "23505" || errcode === "NUNIQ") {
```

benjie (changes requested):
Does this change make sense?

JonParton:
Hmmn would this `or` check need to apply elsewhere then?

@app/client/src/pages/settings/index.tsx
JonParton: Hmmn I'm not totally sure that it can be either?... But if there is the possibility of it being both then there are other areas that will need to account for this that currently only check for `UNIQ` ... it may be better DX to alter the `getCodeFromError(e)` function in [errors.ts](/graphile/starter/blob/main/@app/lib/src/errors.ts#:~:text=export%20function%20getCodeFromError) instead?  Then the Developer can always expect the correct code?

On whether it could be either: 

The `handleErrors` function from [handleErrors.ts](/graphile/starter/blob/main/@app/server/src/utils/handleErrors.ts#:~:text=export%20default%20function%20handleErrors) that intercepts errors and overwrites these codes is registered as the `handleErrors` `PostGraphileOption` in [installPostGraphle.ts](/graphile/starter/blob/main/@app/server/src/middleware/installPostGraphile.ts#:~:text=//%20Custom%20error%20handling,handleErrors%2C)... From reading the [Docs](https://www.graphile.org/postgraphile/usage-library/#:~:text=errcode%27%5D.%20Default%20is%20%5B%5D.-,handleErrors,-%3A%20Enables%20ability%20to) this will run on all errors from the server which the Appolo Client is talking to ... So I thought this would be the only source of this kind of error and so they would always be overwritten??

However I have not delved into the Postgraphile code to confirm that ... can they escape through without getting run through `handleErrors`?

benjie: Ah yes, I forgot we completely overrode all 23505's in handleErrors to be NUNIQ's - good catch :+1:

benjie (approved):
Good catch :+1:

PR #285, opened 2021-12-10
fix: make username sanitization case-insensitive (#284)

### LINKED ISSUE #284: username sanitization is case sensitive; column is citext
### Summary

Creating a new user using oauth fails in case there is already the same username with different case (ie. existing: peter, newly registering: Peter)

### Steps to reproduce

1. create an account with same username as your github username, but in different case (in my case, HYDRANDT would do)
2. sign out
3. attempt to sign in using github
4. enjoy blue screen "An unknown error occurred"

### Expected results

Username sanitization should be case insensitive and username should be sanitized to hydrandt1

### Actual results

HYDRANDT and hydrandt are deemed non identical, number is not appended, and inserting a new user fails, as `username` column on `app_public.users` is type `citext`.

### Additional context

- problem is in the initial mutation, https://github.com/graphile/starter/blob/main/%40app/db/migrations/committed/000001.sql, in `app_private.register_user`:

```
-- Sanitise the username, and make it unique if necessary.
...
  select (
    case
    when i = 0 then v_username
    else v_username || i::text
    end
  ) into v_username from generate_series(0, 1000) i
  where not exists(
    select 1
    from app_public.users
    where users.username = (
      case
      when i = 0 then v_username
      else v_username || i::text
      end
    )
  )
  limit 1;
```

### Possible Solution

Convert username using lower() while checking for uniqueness:

```
  select (
    case
    when i = 0 then v_username
    else v_username || i::text
    end
  ) into v_username from generate_series(0, 1000) i
  where not exists(
    select 1
    from app_public.users
    -- comparing using lowercase to make sure the username is unique (username column is citext -> constraint is not case sensitive)
    where lower(users.username) = (
      case
      when i = 0 then lower(v_username)
      else lower(v_username) || i::text
      end
    )
  )
  limit 1;
```

hydrandt:
## Description

Fixes #284, makes username sanitization case-insensitive to avoid potential conflicts when creating account using oauth providers.

## Performance impact

Should be minimal (using lower() 3x in sanitization loop, potentially could be optimized by only running it once and assigning to a variable, is it worth it?

## Security impact

None.

## Checklist

### DIFF
--- a/@app/db/migrations/committed/000001.sql
+++ b/@app/db/migrations/committed/000001.sql
@@ -1,5 +1,5 @@
 --! Previous: -
---! Hash: sha1:eaf2866060caa0bba319236017c15a40d37a7815
+--! Hash: sha1:fed7686b198e84812a6712d8b4da5add87a5f38f
 
 --! split: 0001-reset.sql
 /*
@@ -1369,10 +1369,11 @@ begin
   where not exists(
     select 1
     from app_public.users
-    where users.username = (
+    -- Compare using lowercase, as username column is case-insensitive
+    where lower(users.username) = (
       case
-      when i = 0 then v_username
-      else v_username || i::text
+      when i = 0 then lower(v_username)
+      else lower(v_username) || i::text
       end
     )
   )
--- a/data/schema.sql
+++ b/data/schema.sql
@@ -2,8 +2,8 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 13.4 (Ubuntu 13.4-0ubuntu0.21.04.1)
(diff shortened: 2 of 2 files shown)


benjie:
Is this actually required? Username is citext so it should already be compared case insensitively…

benjie:
Ah the comparison is text due to concat; we should case the result of concat back to citext.

benjie (changes requested):
Could you try a different approach:

```sql
where users.username = (
  case
  when i = 0 then v_username
  else (v_username || i::text)::citext
  end
)
```

this should solve the issue but also means we can continue to use the username unique index.

PR #307, opened 2022-06-17
Add cypress test for subscription update of user verification

fnimick:
## Description

Adds a test to confirm that the current user subscription used for displaying validation status correctly updates the UI when database modifications are made.

## Performance impact

None.

## Security impact

None.

## Checklist

N/A - this is a test-only change.

### DIFF
--- a/@app/components/src/SharedLayout.tsx
+++ b/@app/components/src/SharedLayout.tsx
@@ -254,7 +254,10 @@ export function SharedLayout({
                   <Avatar>
                     {(data.currentUser.name && data.currentUser.name[0]) || "?"}
                   </Avatar>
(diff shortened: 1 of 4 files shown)


benjie (changes requested):
Awesome; thanks! Are the `cy.wait` commands actually necessary? Cypress automatically waits on an assertion to pass for a certain period of time, perhaps you could just make that period a smidge longer if you need to rather than adding a concrete sleep?

benjie:
Okay that makes sense; it's suboptimal but I doubt it's worth the effort to fix (which could involve having the client write something to window on reconnect and having cypress poll for it or similar strategy) :+1:

PR #315, opened 2022-09-28
Update to postgres 14, node 16

fnimick:
Standardizes on modern versions that new projects should be using.


benjie (approved):
Thanks :raised_hands:

PR #348, opened 2023-03-02
Add types property to @app/lib package.json

fnimick:
Fixes issue where @app/lib cannot be imported in other monorepo packages due to a type resolution error.


benjie (approved):
Great catch!

benjie:
Though I don't know why this doesn't just work by default 🤷‍♂️

PR #349, opened 2023-03-02
Fix docker yarn commands & dockerfile start command

fnimick:
Resolves the following issues:
- `yarn docker start` would run `yarn start` in the top-level directory rather than using the docker-helpers workspace
- The docker container created by `docker compose up` would fail to launch because yarn 3.x requires an initial `yarn` invocation to succeed before any scripts can be run.


benjie:
Would you mind updating the README to indicate that `yarn` needs to be ran before `yarn setup` too please? I meant to do this but forgot.

benjie (approved):
Thanks!

PR #351, opened 2023-03-07
Run a yarn install prior to `yarn setup` call in docker setup

fnimick:
Fixes an issue related to the yarn 3 upgrade where a yarn install is required prior to any yarn scripts being run.


benjie (approved):
Great catch; thanks!

PR #352, opened 2023-03-08
Update yarn-setup.js to also set group

triptec:
When fixing the permissions of node_modules mounted from docker volumes also set the group to node. This does not fix a bug, it works just as it is, but it looks strange especially when debugging and one is looking into every possible problem =).

I also updated the comment around it as at first I could not understand why it would be needed in the first place as I thought the dirs would be owned by node if the node_modules dirs were created by yarn run by the node user.

## Checklist


benjie (approved):
Great; thanks!

PR #353, opened 2023-03-08
Run an `ls` command on container start to fix macOS git perms

fnimick:
Fixes a problem where git commands in the container on macOS failed with a "dubious ownership in repository" error.


benjie (approved):
How utterly bizarre.

The discussion in discord is here: https://discord.com/channels/489127045289476126/681492122184056844/1083069361432051753

Thanks for the fix, even if it doesn't work it looks harmless

PR #355, opened 2023-03-09
Upgrade to PostGraphile V5

benjie:
Note that since PostGraphile V5 is still pre-release sponsors-only software right now, to use this PR you will need to know the secret from https://grafast.org/caveats/ . If you don't have access and are already a sponsor, ping Jem ([on the Graphile Discord](https://discord.gg/graphile), [via Twitter](https://twitter.com/jemgillam), or via email on their first name at graphile.com) and they'll help you get set up.

**IMPORTANT**: when viewing the diff, you should **hide whitespace changes** otherwise it'll look a lot bigger than it really is.

Any questions, ask me in #🔮 on Discord (if you can't see that channel, ping Jem - see above).

NOTE: CI is not passing, because CI is not a sponsor 😉

### DIFF
--- a/@app/__tests__/helpers.ts
+++ b/@app/__tests__/helpers.ts
@@ -1,6 +1,6 @@
 import { Pool, PoolClient } from "pg";
 
-const pools = {};
+const pools: { [key: string]: Pool } = {};
 
 if (!process.env.TEST_DATABASE_URL) {
   throw new Error("Cannot run tests without a TEST_DATABASE_URL");
--- a/@app/server/__tests__/helpers.ts
+++ b/@app/server/__tests__/helpers.ts
@@ -25,8 +25,8 @@ export async function createUserAndLogIn() {
   const pool = poolFromUrl(process.env.TEST_DATABASE_URL!);
   const client = await pool.connect();
   try {
-    const [user] = await createUsers(pool, 1, true);
-    const session = await createSession(pool, user.id);
+    const [user] = await createUsers(client, 1, true);
+    const session = await createSession(client, user.id);
     return { user, session };
   } finally {
     client.release();


@app/lib/src/GraphileApolloLink.ts
```diff
@@ -53,22 +52,32 @@ export class GraphileApolloLink extends ApolloLink {
...
+              node: {
+                req,
+                res,
+              },
+              expressv4: {
+                req,
+                res,
+              },
+            },
+            pgl.getResolvedPreset()
```
benjie, line 74: `hookArgs` is how the context is derived now, it means `withPostGraphileContextFromReqRes` is no longer needed. We pass the `req`/`res` objects (both into `expressv4` and `node` scopes) so that `hookArgs` can use them to call your `context` callback (or `pgSettings`/`additionalGraphQLContextFromRequest` if you're using the V4 preset, which we are) to derive the GraphQL context.

@app/server/src/plugins/PassportLoginPlugin.ts
```diff
@@ -85,12 +87,40 @@ const PassportLoginPlugin = makeExtendSchemaPlugin((build) => {
...
+    },
+    LoginPayload: {
+      user() {
+        const $userId = currentUserIdSource.execute() as PgClassExpressionStep<
+          any,
+          any,
+          any,
+          any,
+          any
+        >;
```
benjie, line 114: The TypeScript types in V5 need improving so that this cast isn't needed 😅

@app/server/src/plugins/PassportLoginPlugin.ts
```diff
@@ -191,29 +208,15 @@ const PassportLoginPlugin = makeExtendSchemaPlugin((build) => {
...
-            (tableAlias, sqlBuilder) => {
-              sqlBuilder.where(
-                sql.fragment`${tableAlias}.id = app_public.current_user_id()`
-              );
-            }
-          );
-          return {
-            data: row,
-          };
+          return {};
```
benjie, line 214: It's now the responsibility of the next layer of plans to execute this. See the `LoginPayload.user` plan above, which is effectively:

```ts
    LoginPayload: {
      user() {
        const $userId = currentUserIdSource.execute();
        return userSource.get({ id: $userId });
      },
    }
```

@app/server/src/plugins/SubscriptionsPlugin.ts
```diff
@@ -67,10 +33,25 @@ const currentUserTopicFromContext = async (
...
+  );
+  if (!currentUserIdSource) {
+    throw new Error("Couldn't find current_user_id source");
+  }
+  const usersSource = build.input.pgSources.find(
+    (s) =>
+      !s.parameters &&
+      s.extensions?.pg?.schemaName === "app_public" &&
+      s.extensions.pg.name === "users"
+  );
```
benjie, line 47: This is an alternative approach if you know the database name rather than the inflected source name. I include it as an alternative to the much simpler equivalent code we saw before:

```ts
const userSource = build.input.pgSources.find((s) => s.name === "users");
```

@app/server/src/plugins/SubscriptionsPlugin.ts
```diff
@@ -79,14 +60,36 @@ const SubscriptionsPlugin = makeExtendSchemaPlugin((build) => {
         """
         Triggered when the logged in user's record is updated in some way.
         """
-        currentUserUpdated: UserSubscriptionPayload @pgSubscription(topic: ${embed(
-          currentUserTopicFromContext
-        )})
+        currentUserUpdated: UserSubscriptionPayload
```
benjie, line 63: The `@pgSubscription` directive is no more, instead write a `subscribePlan` (see below).

srghma:
seems like work on V5 still continues :+1:

PR #357, opened 2023-03-14
Fix `yarn server schema:export` command via makeSchema

jamesallain:
## Description

Fix schema-export. I had to change the script to run it. `cross-env` wasn't working for me. 

## Performance impact

none

## Security impact

unknown

## Checklist


jamesallain:
I think it's failing where `You must set the I_SPONSOR_GRAPHILE environmental variable to the shared secret to use this pre-release software.`

benjie (approved):
Thanks for catching this, apparently this file doesn't get type checked with everything else!

PR #359, opened 2023-03-16
Explicitly connect redis-client before providing to connect-redis

fnimick:
Required as of redis v4.x and connect-redis v7.x

Resolves a `ClientClosedError` that occurs on requests if a REDIS_URL is provided.


benjie (approved):
Thanks for catching this!

